PR #1, opened 2020-06-28
Add GitHub CI config

christophebedard:
This PR configures a GitHub workflow that matches the existing GitLab CI: https://gitlab.com/ros-tracing/ros2_tracing/-/blob/b695e4e367fb6e250d3147ab6b851b25a0e50299/.gitlab-ci.yml. It builds `ros2_tracing` with/without LTTng and with/without tracing being disabled, and checks if the tracing instrumentation is indeed available (or not).

The "sanity checks" workflow makes sure that the tracing instrumentation can be enabled when starting from a default source build or a binary installation, as explained in the build instructions in the README: https://github.com/ros2/ros2_tracing#building. This will also be documented on docs.ros.org, see: https://github.com/ros2/ros2_documentation/pull/2877.

### DIFF
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -1,0 +1,21 @@
+name: DCO
+on:
+  pull_request:
+  push:
+    branches:
+      - rolling
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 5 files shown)


christophebedard:
This only affects local CI, so I'll merge this.

PR #6, opened 2022-07-30
Only run DCO check job for PRs

christophebedard:


christophebedard:
This only affects the DCO check job, so skipping all other CI jobs.

PR #7, opened 2022-08-12
Use binaries from pre-release/testing apt repository

christophebedard:
Otherwise binary jobs can fail if breaking-ish changes are introduced, since we still build part of the code from source.


christophebedard:
This doesn't actually fix the current CI failures, but the change definitely makes sense, so we'll go ahead.

PR #33, opened 2023-01-31
Update tracing to C++17.

clalancette:
The main reason to do this is so that we can compile tracing with the clang static analyzer.  As of clang++-14 (what is in Ubuntu 22.04), the default still seems to be C++14, so we need to specify C++17 so that new things in the rclcpp headers work properly.

Further, due to reasons I don't fully understand, I needed to set CMAKE_CXX_STANDARD_REQUIRED in order for clang to really use that version.  So set this as well.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/test_tracetools/CMakeLists.txt
+++ b/test_tracetools/CMakeLists.txt
@@ -1,9 +1,10 @@
 cmake_minimum_required(VERSION 3.5)
 project(test_tracetools)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")


test_tracetools/CMakeLists.txt
```diff
@@ -1,9 +1,10 @@
 cmake_minimum_required(VERSION 3.5)
 project(test_tracetools)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
christophebedard, line 7: I'm guessing making the same changes in `tracetools` isn't strictly necessary? It would probably be a good idea, though.

clalancette: Oh, you know, it didn't come up in my compilation tests with clang, so I didn't include it.  But I agree that we should just make it consistent, so I'll update this PR with that.

clalancette: Done in 3d3a885

tracetools/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
 cmake_minimum_required(VERSION 3.5)
 project(tracetools)
 
-# Default to C++14
+# Default to C++17
 if(NOT CMAKE_CXX_STANDARD)
-  set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 17)
```
christophebedard, line 6: I'm assuming you need this here too

```suggestion
  set(CMAKE_CXX_STANDARD 17)
  set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

clalancette: Yes, of course, thank you.

christophebedard (approved):

clalancette:
CI is in https://github.com/ros2/rcl/pull/1031#issuecomment-1412130409 .  That was green, so merging.

PR #43, opened 2023-02-14
Fix memory leak in tracetools::get_symbol()

### LINKED ISSUE #34: Memory leak with tracetools::get_symbol()
`tracetools::get_symbol()` calls `tracetools::detail::demangle_symbol()`, which calls `abi::__cxa_demangle()`. The returned pointer [should be freed by the caller](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.3/a01696.html), but it isn't.

`tracetools::get_symbol()` is only called when the `rclcpp_callback_register` tracepoint is called, so this only happens during initialization, and the size of the leaked memory is proportional to the number of callbacks (timers, subscriptions, services). This should still be fixed before https://github.com/ros2/ros2/issues/1177, though.

Working on a fix.

christophebedard:
Fixes #34

`tracetools::get_symbol()` is used to get a demangled symbol from a function pointer or `std::function` object. This is used to associate a human-readable string to a callback in the trace data. To do symbol demangling, it calls [`abi::__cxa_demangle()`, which returns a pointer that must be freed by the caller](https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.3/a01696.html). This was never freed. The best solution would be to use a limited size buffer on the stack and never allocate. However, while `abi::__cxa_demangle` _can_ accept a pre-allocated buffer, it cannot be on the stack, since `abi::__cxa_demangle` expands it using `realloc` if it is not long enough.

Since the pointer to the demangled symbol must be freed, the problem is that `tracetools::get_symbol()` was returning a pointer to a literal when something failed (e.g., demangling, or mapping from function pointer to shared object), so we couldn't always `free()` the return value. There's no real use to returning anything other than `nullptr` when something fails.

Changes:

* Change `tracetools::get_symbol()` to return `char *` instead of `const char *`, and to return `nullptr` when something fails
   * Do the same for the underlying/internal functions.
   * Document that the return value should be freed by the caller.
* Change `get_symbol_funcptr(void * funcptr)` to `get_symbol_funcptr(const void * funcptr)`, since that's more appropriate
* Remove the `#ifdef TRACETOOLS_LTTNG_ENABLED` conditions in `tracetools::detail::{demangle_symbol,get_symbol_funcptr}`, since there's no real need to do anything different if `TRACETOOLS_LTTNG_ENABLED` is defined. These functions shouldn't be called in that case anyway.
* Update corresponding test (`test_tracetools`'s `test_utils`)

This PR breaks ABI and requires the following downstream changes:

* Update calls to `tracetools::get_symbol()` to free the memory
   * Downstream callers could use the new `TRACEPOINT_ENABLED()` and `DO_TRACEPOINT()` macros (#46) to avoid allocating memory when the tracepoint is disabled

Downstream PR:

* https://github.com/ros2/rclcpp/pull/2104

### DIFF
--- a/test_tracetools/test/test_utils.cpp
+++ b/test_tracetools/test/test_utils.cpp
@@ -37,54 +37,62 @@ public:
   }
 };
 
-/*
-   Testing symbol resolution for std::function object created from a function pointer.
+/**
+ * Testing symbol resolution for std::function object created from a function pointer.
  */
 TEST(TestUtils, valid_symbol_funcptr) {
   std::function<void(std::shared_ptr<int>)> f = &function_shared;
-  EXPECT_STREQ(tracetools::get_symbol(f), "function_shared(std::shared_ptr<int>)") <<
+  char * symbol = tracetools::get_symbol(f);
+  EXPECT_STREQ(symbol, "function_shared(std::shared_ptr<int>)") <<
     "invalid symbol";
+  std::free(symbol);
 }
 
-/*
-   Testing get_symbol_funcptr from a null pointer.
+/**
+ * Testing get_symbol_funcptr from a null pointer.
  */
 TEST(TestUtils, invalid_get_symbol_funcptr) {
-  EXPECT_STREQ(tracetools::detail::get_symbol_funcptr(nullptr), TRACETOOLS_SYMBOL_UNKNOWN);
+  EXPECT_EQ(tracetools::detail::get_symbol_funcptr(nullptr), nullptr);
 }
 
-/*
-   Testing symbol resolution for std::function object created from a lambda.
+/**
+ * Testing symbol resolution for std::function object created from a lambda.
  */
 TEST(TestUtils, valid_symbol_lambda) {
   std::function<int(int)> l = [](int num) {return num + 1;};
+  char * symbol = tracetools::get_symbol(l);
   EXPECT_STREQ(
-    tracetools::get_symbol(l),
+    symbol,
     "TestUtils_valid_symbol_lambda_Test::TestBody()::{lambda(int)#1}") <<
     "invalid symbol";
+  std::free(symbol);
 }
 
-/*
-   Testing symbol resolution lambdas with capture.
+/**
+ * Testing symbol resolution lambdas with capture.
  */
 TEST(TestUtils, valid_symbol_lambda_capture) {
   int num = 1;
 
   auto l = [ = ]() {return num + 1;};
+  char * symbol = tracetools::get_symbol(l);
   EXPECT_STREQ(
-    tracetools::get_symbol(l),
+    symbol,
     "TestUtils_valid_symbol_lambda_capture_Test::TestBody()::{lambda()#1}") <<
     "invalid symbol";
+  std::free(symbol);
 
   auto m = [&](int other_num) {return num + other_num;};
+  symbol = tracetools::get_symbol(m);
   EXPECT_STREQ(
-    tracetools::get_symbol(m),
+    symbol,
     "TestUtils_valid_symbol_lambda_capture_Test::TestBody()::{lambda(int)#2}") <<
     "invalid symbol";
+  std::free(symbol);
 }
 
-/*
-   Testing symbol resolution for std::function object created from std::bind.
+/**
+ * Testing symbol resolution for std::function object created from std::bind.
  */
 TEST(TestUtils, valid_symbol_bind) {
   SomeClassWithCallback scwc;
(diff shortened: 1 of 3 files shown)


christophebedard:
I'm marking this as a draft because I would like to get feedback before updating downstream packages.

This PR requires the following change in `rclcpp`, everywhere `TRACEPOINT(rclcpp_callback_register, ...)` is called:

```diff
+ char * symbol = tracetools::get_symbol(callback_);
  TRACEPOINT(
    rclcpp_callback_register,
    static_cast<const void *>(&callback_),
-   tracetools::get_symbol(callback_));
+   symbol);
+ std::free(symbol);
```

Of course, we can surround this with `#ifndef TRACETOOLS_DISABLED`/`#endif` to avoid calling `tracetools::get_symbol()` and allocating memory if the instrumentation is completely removed. However, if the instrumentation is included, then, even if the tracepoint is disabled (i.e., even if the data isn't actually collected by the tracer), memory will be allocated.

To solve this, we can replace LTTng's `tracepoint(...)` with:

```cpp
if (tracepoint_enabled(name)) {
  // Compute arguments
  do_tracepoint(name, ...);
}
```
See [lttng-ust(3)](https://lttng.org/man/3/lttng-ust/v2.13/). Then we can first check if the tracepoint is enabled before calling `tracetools::get_symbol()`/allocating, triggering the tracepoint manually, and freeing the memory. However, this would mean having to create the corresponding macros to go [alongside the current `TRACEPOINT(...)` macro](https://github.com/ros2/ros2_tracing/blob/da37d6a7e6ea8376707f4ee453618277ed7e98d5/tracetools/include/tracetools/tracetools.h#L78), i.e., `TRACEPOINT_ENABLED(...)` and `DO_TRACEPOINT(...)`. This complicates `tracetools` a bit, but it allows for more flexibility for things like this.

christophebedard:
cc @iluetkeb @mjcarroll

iluetkeb:
Good catch!

The current approach basically passes on the obligation to free the memory to the caller of get_symbol. This is the most straightforward approach, but I'm wondering whether we can come up with something to avoid it. You have thought about the pre-allocated buffer (but we would also have to free _that_ somehow).

btw, one of the other thoughts I had is whether we could output metadata throughout execution at regular intervals. This would then _require_ us to store the metadata anyhow. I'm not saying we need that, but it's maybe something to consider in conjunction.

christophebedard:
> The current approach basically passes on the obligation to free the memory to the caller of get_symbol. This is the most straightforward approach, but I'm wondering whether we can come up with something to avoid it. You have thought about the pre-allocated buffer (but we would also have to free _that_ somehow).

`demangle_symbol()` could have its own buffer (on the stack), and:

1. Call `abi::__cxa_demangle`, let it allocate
2. Copy the result into its buffer, up to some max length
   * I could come up with a reasonable value; in any case this isn't really critical
3. Free the string returned by `abi::__cxa_demangle`
4. Return a `const char *` to the buffer

This wouldn't really work if you called `tracetools::get_symbol()` more than once before the tracer can read the first string, but we might not need to call it consecutively anyway, and we could document this. This makes it easier for the caller, but this might be too weird of a behaviour. As you said, the simpler approach is to let the caller free the memory.

christophebedard:
> btw, one of the other thoughts I had is whether we could output metadata throughout execution at regular intervals. This would then _require_ us to store the metadata anyhow. I'm not saying we need that, but it's maybe something to consider in conjunction.

I opened an issue for this an mentioned that as a potential solution: #44.

christophebedard:
> However, this would mean having to create the corresponding macros to go [alongside the current `TRACEPOINT(...)` macro](https://github.com/ros2/ros2_tracing/blob/da37d6a7e6ea8376707f4ee453618277ed7e98d5/tracetools/include/tracetools/tracetools.h#L78), i.e., `TRACEPOINT_ENABLED(...)` and `DO_TRACEPOINT(...)`. This complicates `tracetools` a bit, but it allows for more flexibility for things like this.

Note that this improvement in `tracetools` (and downstream changes in `rclcpp`) could be made in separate PRs.

christophebedard:
Testing `--packages-above tracetools test_tracetools rclcpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18089/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12621/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18733/)

Includes:

* This PR
* https://github.com/ros2/rclcpp/pull/2104

christophebedard:
@mjcarroll friendly ping

mjcarroll (approved):

cottsay:
I think this caused a fair number of regressions. This repository's CI shows breakage, and my builds on RHEL 8 are now failing as well.

Stuff like:
```
libtracetools.so: undefined reference to `dladdr'
collect2: error: ld returned 1 exit status
```

christophebedard:
> I think this caused a fair number of regressions. This repository's CI shows breakage, and my builds on RHEL 8 are now failing as well.
> 
> 
> 
> Stuff like:
> 
> ```
> 
> libtracetools.so: undefined reference to `dladdr'
> 
> collect2: error: ld returned 1 exit status
> 
> ```

Are there any non-RHEL failures?

christophebedard:
Looking at CI, it seems to be RHEL-only. #48 fixes that.

PR #47, opened 2023-03-01
Revert "Fix memory leak in tracetools::get_symbol()"

mjcarroll:
Looks like this introduced regressions.

Reverts ros2/ros2_tracing#43


christophebedard:
We won't need to revert (see #48), so closing this.

PR #48, opened 2023-03-01
Explicitly link against dl for dladdr()

christophebedard:
Fix regression introduced by #43 on RHEL.

It looks like we need to explicitly link against dl on RHEL. Note that, if `TRACETOOLS_LTTNG_ENABLED`, LTTng's link flags already include `-ldl`.

### DIFF
--- a/tracetools/CMakeLists.txt
+++ b/tracetools/CMakeLists.txt
@@ -81,6 +81,10 @@ if(TRACETOOLS_LTTNG_ENABLED)
     target_link_libraries(${PROJECT_NAME} "-rdynamic")
   endif()
 endif()
+if(NOT TRACETOOLS_DISABLED)
+  # Explicitly linking against dl for dladdr() seems to be required for RHEL
+  target_link_libraries(${PROJECT_NAME} "${CMAKE_LINK_LIBRARY_FLAG}dl")
+endif()
 
 if(NOT TRACETOOLS_DISABLED)
   # Only use output/binary include directory


christophebedard:
CI job showing `tracetools` build failure on RHEL: https://ci.ros2.org/job/ci_linux-rhel/322/consoleFull#console-section-217

A different but functionally identical commit makes `tracetools` build fine on RHEL: https://ci.ros2.org/job/ci_linux-rhel/323/console#console-section-217

christophebedard:
Testing `--packages-above tracetools`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18127/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12661/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18777/)

and (same as the one I linked to above):

* RHEL [Build Status](https://ci.ros2.org/job/ci_linux-rhel/323/)

mjcarroll:
@cottsay would appreciate a RHEL-specific set of eyes on this one, thanks!

mjcarroll (approved):

tracetools/CMakeLists.txt
```diff
@@ -80,6 +80,10 @@ if(TRACETOOLS_LTTNG_ENABLED)
   if(NOT TRACETOOLS_NO_RDYNAMIC)
     target_link_libraries(${PROJECT_NAME} "-rdynamic")
   endif()
+elseif(NOT WIN32)
+  # Explicitly linking against dl for dladdr() seems to be required for RHEL
+  # (LTTng's link flags already include it above)
+  target_link_libraries(${PROJECT_NAME} "-ldl")
 endif()
```
cottsay, line 87: Thanks for finding a fix so quickly. I'm happy to see that this didn't take out the Ubuntu builds like I initially thought it did.

This is probably fine as-is, but I have two soft suggestions:
1. This is an opportunity to employ a more defensive programming technique with regard to external dependencies changing their projects. I'm not sure it's a good idea to rely on `lttng-ust` to provide a library that's a direct dependency of this one.
2. Use the `CMAKE_LINK_LIBRARY_FLAG` macro instead of verbatim `-l`.

```suggestion
if(NOT WIN32)
  # Explicitly linking against dl for dladdr() seems to be required for RHEL
  # (LTTng's link flags already include it above)
  target_link_libraries(${PROJECT_NAME} "${CMAKE_LINK_LIBRARY_FLAG}dl")
endif()
```

cottsay (approved):

tracetools/CMakeLists.txt
cottsay: Another thing to note here is that the reason RHEL hit this is that we're not installing the `lttng-ust` development package in our CI at all. I took a look, and it it would have been present, it would have provided the `-ldl` link flag. Interestingly, that dependency isn't declared in this package's manifest.

The reason we're not installing `lttng-ust-devel` is that the system package version is too old for this package to use. I vaguely remember opening an issue or making a PR a while ago to improve the dependency version detection to reject the older package, but I can't recall where that landed.

cottsay: Another interesting note is that our nightly CI builds on [build.ros2.org](https://build.ros2.org/view/Rci/job/Rci__nightly-release_ubuntu_jammy_amd64/369) install dependencies using rosdep and therefore didn't install the `lttng-ust` development package, but it didn't cause the same problem. Maybe Ubuntu links in `dl` implicitly?

christophebedard: > This is probably fine as-is, but I have two soft suggestions:

Thanks!

> Another thing to note here is that the reason RHEL hit this is that we're not installing the `lttng-ust` development package in our CI at all. I took a look, and it it would have been present, it would have provided the `-ldl` link flag.

That's correct, but the reason this broke after #43 is because we were `#ifdef`ing the `dladdr()` call out if not `TRACETOOLS_LTTNG_ENABLED`.

> Interestingly, that dependency isn't declared in this package's manifest.
> The reason we're not installing lttng-ust-devel is that the system package version is too old for this package to use. I vaguely remember opening an issue or making a PR a while ago to improve the dependency version detection to reject the older package, but I can't recall where that landed.

I think it was this issue: https://github.com/ros2/ros2_tracing/issues/16 (or the original GitLab issue). However, I think you recently said that RHEL has LTTng 2.12, which would be fine (unlike 2.8 like you mentioned in the original GitLab issue).

We intentionally do not declare a dependency on LTTng-UST, but we plan on changing that soon, see https://github.com/ros2/ros2/issues/1177.

christophebedard: > Maybe Ubuntu links in dl implicitly?

Yeah, it would seem so.

cottsay: > I think you recently said that RHEL has LTTng 2.12

Yes, RHEL 9 has LTTng 2.12, but we haven't quite switched Rolling over from RHEL 8 yet, which has 2.8.1. It would probably be a good idea to follow through with #16 prior to deploying ros2/ros2#1177 even after we get Rolling onto RHEL 9 - I think it's safe to say that older versions are not uncommon in the wild.

cottsay:
Fresh CI:
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/18135/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/12673/)
* Linux-rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/326/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/18785/)

tracetools/CMakeLists.txt
```diff
@@ -80,6 +80,9 @@ if(TRACETOOLS_LTTNG_ENABLED)
   if(NOT TRACETOOLS_NO_RDYNAMIC)
     target_link_libraries(${PROJECT_NAME} "-rdynamic")
   endif()
+if(NOT WIN32)
```
cottsay, line 83: Gah, this is my fault. I didn't test this change before I suggested it.

```suggestion
endif()
if(NOT WIN32)
```

christophebedard:
> * Linux-rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/326/)

FYI that job is testing a different branch (my initial fix PoC). I've deleted it now.

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18139/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12677/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/327/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18787/)

tracetools/CMakeLists.txt
```diff
@@ -81,6 +81,10 @@ if(TRACETOOLS_LTTNG_ENABLED)
     target_link_libraries(${PROJECT_NAME} "-rdynamic")
   endif()
 endif()
+if(NOT WIN32)
```
christophebedard, line 84: Haven't been able to get Windows CI to run yet, but I think it will fail. I think this should be:

```suggestion
if(NOT TRACETOOLS_DISABLED AND NOT WIN32)
```

If `TRACETOOLS_DISABLED`, `tracetools` is only an interface library, and so `target_link_libraries()` must be called with `INTERFACE`. We don't really need to link against `dl` in that case anyway, so we should just avoid doing this if `TRACETOOLS_DISABLED`. @cottsay thoughts?

cottsay: > Haven't been able to get Windows CI to run yet, but I think it will fail.

If it passed before, why would it fail with this PR? This block is only processed on non-Windows platforms.

> If `TRACETOOLS_DISABLED`, tracetools is only an interface library, and so `target_link_libraries()` must be called with `INTERFACE`.

I think you're right. We can test it with a quick Linux CI build though: [Build Status](https://ci.ros2.org/job/ci_linux/18141/)

christophebedard: > If it passed before, why would it fail with this PR? This block is only processed on non-Windows platforms.

Yeah, I quickly realized and edited my comment :sweat_smile:

christophebedard: Alright, I changed it to `if(NOT TRACETOOLS_DISABLED)`. I'll trigger CI again.

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18142/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12681/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/328/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18802/)

tracetools/CMakeLists.txt
christophebedard: > We can test it with a quick Linux CI build though: [Build Status](https://ci.ros2.org/job/ci_linux/18141/)

This should've been `-DTRACETOOLS_DISABLED:BOOL=ON` instead of `-DTRACETOOLS_DISABLED:BOOL=OFF` to test this, but yeah we don't really need to test this.

cottsay: :facepalm: I'll cancel it.

clalancette:
The test failures on aarch64 and RHEL are known, and match the nightlies.  The Windows test failures are new, but I have a hard time seeing how this PR would cause them given its contents.  I'm going to go ahead and merge this (and then do a new release here), and we'll see what happens.

PR #49, opened 2023-03-02
5.1.0

clalancette:
This PR releases 5.1.0.

Note that I chose a minor bump because the set of changes since 5.0.0 includes the addition of an API.

Once this is approved and merged, I will tag and release the package for Rolling.


mjcarroll:
> This PR releases 5.0.0.

5.1.0?

> Note that I chose a minor bump because the set of changes since 5.0.0 includes the addition of an API.

I think technically we broke ABI here, but the 5.0.0 release just happened, so minor may be okay.

clalancette:
> 5.1.0?

Yes, sorry.  Typo, fixed now.

> I think technically we broke ABI here, but the 5.0.0 release just happened, so minor may be okay.

I haven't typically been bumping major for ABI breaks (only for API breaks), but I don't feel strongly either way.

mjcarroll (approved):

PR #54, opened 2023-04-06
Improve 'ros2 trace' command error handling & add end-to-end tests

### LINKED ISSUE #26: Add end-to-end test for 'ros2 trace' command
(Issue moved from GitLab, see: https://gitlab.com/ros-tracing/ros2_tracing/-/issues/136)

### LINKED ISSUE #36: Destroy tracing session if it exists when tracing configuration fails
`ros2 trace` creates a tracing session, then tries to enable userspace & kernel events and add context field types. If something in there fails, then the tracing session still exists. Having something linger on in case of a failure is annoying.

For example, when enabling a context type that isn't recognized:

```
$ ros2 trace -l -c 'unknown' -s test
UST tracing enabled (52 events)
       ...
kernel tracing disabled
context (1 names)
        unknown
writing tracing session to: /home/$USER/.ros/tracing/test
press enter to start...
failed to find context type: unknown
$ lttng list
Available recording sessions:
  1) test [inactive]
    Trace output: /home/$USER/.ros/tracing/test

Use lttng list <session_name> for more details
```

The user must then destroy the tracing session:

```
$ lttng destroy test
Session test destroyed
```

The solution could be to simply destroy the tracing session if an exception is thrown.

When creating the tracing session, we do destroy it if it already exists (see `_create_session()`), so _not_ destroying the session will not cause any issues if the user tries running `ros2 trace` with the same session name, but it's cleaner to clean up and destroy the session after a failure.

We should check if this fix also applies to the `Trace` launch action; I think it should, since this is all in the `setup()` function.

Finally, a test would be great here, e.g., set up a tracing session so that it fails and assert that the tracing session doesn't exist.

christophebedard:
Closes #26

Closes #36

This improves error handling and error report in `tracetools_trace`. The common functions used by `ros2trace` (for the `ros2 trace` command) and `tracetools_launch` (for the `Trace` launch action) now do not print anything; they raise exceptions where appropriate. Then, `ros2 trace`/`Trace` can catch any error and report it using the appropriate method: print to `stderr` for `ros2 trace` and log error for `Trace`.

This also makes sure to destroy the tracing session if it exists when tracing configuration fails.

Finally, this adds end-to-end tests for the `ros2 trace` command to cover the main cases. It also improves parameter validation for the session name.

### DIFF
--- a/ros2trace/package.xml
+++ b/ros2trace/package.xml
@@ -20,7 +20,13 @@
   <test_depend>ament_mypy</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>ament_xmllint</test_depend>
+  <test_depend>launch</test_depend>
+  <test_depend>launch_ros</test_depend>
   <test_depend>python3-pytest</test_depend>
+  <test_depend>ros2run</test_depend>
+  <test_depend>test_tracetools</test_depend>
+  <test_depend>tracetools</test_depend>
+  <test_depend>tracetools_read</test_depend>
 
   <export>
     <build_type>ament_python</build_type>
--- a/ros2trace/ros2trace/command/trace.py
+++ b/ros2trace/ros2trace/command/trace.py
@@ -16,8 +16,7 @@
 
 from ros2cli.command import CommandExtension
 from tracetools_trace.tools import args
-from tracetools_trace.trace import fini
-from tracetools_trace.trace import init
+from tracetools_trace.trace import trace
 
 
 class TraceCommand(CommandExtension):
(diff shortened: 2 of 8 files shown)


christophebedard:
Hmm maybe these tests shouldn't rely on tracepoints being included. Or maybe they could be skipped if tracepoints are not included.

christophebedard:
> Or maybe they could be skipped if tracepoints are not included.

I did this.

christophebedard:
Testing `--packages-above ros2trace tracetools_launch tracetools_trace`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18584/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13126/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19321/)

mjcarroll (approved):

PR #55, opened 2023-04-08
Update README and other documentation

christophebedard:
* Delete `.gitlab-ci.yml`, since we don't use GitLab (CI) anymore
* Delete `instrumented.repos`, since we don't need to build any of the instrumented core repos from source anymore
* General updates to the README and quality declaration documents, including:
   * Don't say that `ros2 trace` will enable "a few kernel tracepoints" by defaults, since it no longer enables kernel tracing by default


aprotyas (approved):

christophebedard:
This only changes documentation/non-code files, so I'm skipping CI.

PR #59, opened 2023-04-12
Allow requiring minimum lttng package version for is_lttng_installed

### LINKED ISSUE #16: Gracefully handle older versions of LTTng

christophebedard:
Closes #16

This changes `is_lttng_installed()` to also check for the `lttng` Python package.

This also adds an optional `minimum_version` parameter to `is_lttng_installed()` to check if the `lttng` Python package version is sufficient. This is then used to ignore some tests which require `lttng.BUFFER_*`, which isn't available until version 2.9. RHEL 8 has LTTng 2.8, so those tests fail. With this change, tests run successfully on RHEL 8, or, rather, the tests that can't run are skipped.

### DIFF
--- a/test_tracetools/test/test_executor.py
+++ b/test_tracetools/test/test_executor.py
@@ -16,8 +16,10 @@ import unittest
 
 from tracetools_test.case import TraceTestCase
 from tracetools_trace.tools import tracepoints as tp
+from tracetools_trace.tools.lttng import is_lttng_installed
 
 
+@unittest.skipIf(not is_lttng_installed(minimum_version='2.9.0'), 'LTTng is required')
 class TestExecutor(TraceTestCase):
 
     def __init__(self, *args) -> None:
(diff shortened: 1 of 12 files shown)


christophebedard:
I should actually look into whether tracing can still be done with 2.8; perhaps we're just not able to configure the buffer type. That wouldn't really affect the tests. Then I could remove the >=2.9.0 requirements.

cottsay (approved):
Thank you very much for making this happen.

It appears to work as expected: [Build Status](https://ci.ros2.org/job/ci_linux-rhel/391/)

christophebedard:
> I should actually look into whether tracing can still be done with 2.8; perhaps we're just not able to configure the buffer type. That wouldn't really affect the tests. Then I could remove the >=2.9.0 requirements.

I can look into this later, this works for now.

Testing `test_tracetools`, `tracetools_trace`, `test_tracetools_launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18523/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13069/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19253/)
* Linux-rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/392/)

PR #65, opened 2023-04-19
Expect binaries to include tracepoints in sanity check CI job

christophebedard:
Follow-up to #31 now that the Rolling binaries include tracepoints.


christophebedard:
We can skip ci.ros2.org here since this only modifies the local GitHub CI config.

mjcarroll (approved):

PR #66, opened 2023-04-20
Detect issue with LTTng and Docker and report error when tracing

christophebedard:
There is an issue with LTTng when using more than one Docker container that share the same mounted home directory. When trying to trace in a second container, LTTng will mistakenly _think_ that there is a session daemon running, so it will happily trace, but no trace data will be recorded, because there isn't actually a session daemon in that container. The fix is to use a different LTTng home, e.g., by setting the `LTTNG_HOME` environment variable to a unique value. See this LTTng issue: https://bugs.lttng.org/issues/1371

This PR detects this (very small) corner case and reports the issue to the user so that they can set `LTTNG_HOME`. Example:

1. Tracing works fine in the first container:

```
container1$ ros2 trace -s my-sesh-container1
UST tracing enabled (59 events)
kernel tracing disabled
context (3 fields)
writing tracing session to: /home/user/.ros/tracing/my-sesh-container1
press enter to start...
press enter to stop...
stopping & destroying tracing session
```

2. But an error will be reported in the second container:

```
container2$ ros2 trace -s my-sesh-container2
UST tracing enabled (59 events)
kernel tracing disabled
context (3 fields)
writing tracing session to: /home/user/.ros/tracing/my-sesh-container2
press enter to start...
error: lttng-sessiond seems to exist, but is unreachable. If using two containers with the same HOME directory, set the LTTNG_HOME environment variable to a unique value for each container. See: https://bugs.lttng.org/issues/1371
```

3. Setting `LTTNG_HOME` to a unique value (so that container 2 has its own LTTng home) and tracing again will work:

```
container2$ export LTTNG_HOME="$HOME/.lttng_home2"
container2$ ros2 trace -s my-sesh-container2
UST tracing enabled (59 events)
kernel tracing disabled
context (3 fields)
writing tracing session to: /home/user/.ros/tracing/my-sesh-container2
press enter to start...
press enter to stop...
stopping & destroying tracing session
```


mjcarroll (approved):
A small corner case, but worth covering as Docker is pretty popular in our community.

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18660/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13188/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19388/)

Issue #44, opened 2023-02-14
Find solution to allow starting to trace from any point in time


christophebedard:
As mentioned in #40 (and [in the README](https://github.com/ros2/ros2_tracing#tracing)), we need to start tracing before the application launches to collect initialization information/metadata, otherwise the tracing data from the "runtime phase" (e.g., message publication data) is meaningless.

The current workaround is to start tracing before the application launches to collect the initialization data, then stop after 30-60 s, and then later on (when needed) start tracing again. This still requires us to actually set up tracing before the application launches. We would ideally like to be able to decide to enable tracing while an application is already running.

Some ideas:

1. Store initialization metadata and re-emit it at regular intervals. Then we'd simply need to record one metadata "burst." This does not sound very efficient, though.
    - This probably requires a bit of actual logic/code in `tracetools`.
2. Store initialization metadata (e.g., in simple caches) and "dump" it when we start tracing, which is similar to what LTTng does for its own stuff.
    - Look into LTTng "statedumps," which provide a dump of the current/initial state of the system so that a trace viewer/processor can make sense of the normal trace data that follows. 
    - This probably requires a bit of actual logic/code in `tracetools`.
3. Always trace the start of an application until some point (e.g., a function called by the user to declare that initialization is now over), possibly leveraging LTTng's [snapshot mode](https://lttng.org/docs/v2.13/#doc-taking-a-snapshot) to avoid writing the trace data to disk if it doesn't end up being needed (if we do not end up enabling tracing later).

This all would make live tracing more useful, so this is related to #149.

ShravanDeva5327:
I am working on this as a part of [Improvements to ROS 2 Tracing](https://summerofcode.withgoogle.com/programs/2025/projects/quKpaaVA)

> Note: This is the initial workflow that @christophebedard and I have planned, we'll refine and update it as the project progresses.

**0. Defining the use case and requirements**

Situation we are trying to address:

- The user launches an application built with tracing instrumentation (which is enabled by default in ROS 2 now), but does not start tracing (neither with `ros2 trace` nor LTTng commands) before the application starts.
- Some time passes, that is, the initialization phase of ROS 2 completes.
- Eventually, the user wants to start tracing. This could be either:
a. To trace upcoming events, or
b. Because they notice something went wrong, in which case, they would also need tracing data from 3-4 minutes before the tracing was started (LTTng's flight recorder mode might help in this case).

Get input from other ROS 2 users and understand potential additional use cases of this feature.
[...]

PR #70, opened 2023-04-24
Create start/pause/resume/stop sub-commands for 'ros2 trace'

### LINKED ISSUE #68: Provide non-interactive 'ros2 trace start/pause/resume/stop' commands
Currently, `ros2 trace` cannot be used in scripts or in non-interactive environments. After running the `ros2 trace` command, the user needs to press start to start tracing and then enter again to stop tracing. We could provide `start` and `stop` sub-commands to do that instead. Also, sub-commands like `pause` and `resume` could help with #44:

```
$ ros2 trace start  ... # Same args as 'ros2 trace' except for a mandatory explicit session name, but starts tracing and returns immediately
# ... wait until initialization is done
$ ros2 trace pause <session_name>  # Pauses tracing and returns immediately
# ... some time later
$ ros2 trace resume <session_name>  # Starts tracing again and returns immediately
$ ros2 trace stop <session_name>  # Stops tracing and returns immediately
```

Here is a comparison with the `lttng` commands:

* `ros2 trace start ...`: equivalent to configuring the tracing session (session name, events, etc.) and `lttng start`
* `ros2 trace pause <session_name>`: equivalent to `lttng stop`
* `ros2 trace resume <session_name>`: equivalent to `lttng start`
* `ros2 trace stop <session_name>`: equivalent to `[lttng stop &&] lttng destroy`

We could offer a `ros2 trace create ...`, which would configure the tracing session _without_ starting to trace, i.e., `ros2 trace start ...` without `lttng start`. However, since there's currently no point in configuring the tracing session without also starting to trace, this wouldn't really be useful. We can always add it later on anyway.

The implementation wouldn't be too bad. We already have separate `setup()`, `start()`, `stop()`, and `destroy()` functions in `tracetools_trace/tracetools_trace/tools/lttng_impl.py`.

Finally, note that `lttng`'s `start`/`stop`/`destroy` don't require a session name argument, since the "current" session name is written to `~/.lttngrc`: https://lttng.org/man/7/lttng-concepts/v2.13/#doc-_current_recording_session. This isn't the case when tracing using the LTTng Python bindings, but we could re-implement this functionality (e.g., write the current session name to `~/.ros/tracing/.rc`).

christophebedard:
Closes #68

That adds the following sub-commands to `ros2 trace` as "verbs":

* `ros2 trace start`: configure tracing session and start tracing
* `ros2 trace pause`: stop tracing, without removing tracing session
* `ros2 trace resume`: start tracing again
* `ros2 trace stop`: stop tracing and destroy tracing session

The goal is to provide non-interactive & finer-grained ways to control tracing. For example, this could be used in scripts. It could also be used for long-term tracing (related to #44): users can trace the first 10-30 seconds to collect initialization information, then pause, and then resume tracing again later on.

Example:

```
$ ros2 trace start my_session
UST tracing enabled (35 events)
kernel tracing disabled
context (3 fields)
writing tracing session to: /home/user/.ros/tracing/my_session
$ ros2 trace pause my_session
$ ros2 trace resume my_session
$ ros2 trace stop my_session
```

### DIFF
--- a/ros2trace/ros2trace/verb/__init__.py
+++ b/ros2trace/ros2trace/verb/__init__.py
@@ -1,0 +1,44 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2cli.plugin_system import PLUGIN_SYSTEM_VERSION
+from ros2cli.plugin_system import satisfies_version
+
+
+class VerbExtension:
+    """
+    The extension point for 'trace' verb extensions.
+
+    The following properties must be defined:
+    * `NAME` (will be set to the entry point name)
+
+    The following methods must be defined:
+    * `main`
+
+    The following methods can be defined:
+    * `add_arguments`
+    """
+
+    NAME = None
+    EXTENSION_POINT_VERSION = '0.1'
+
+    def __init__(self):
+        super(VerbExtension, self).__init__()
+        satisfies_version(PLUGIN_SYSTEM_VERSION, '^0.1')
+
+    def add_arguments(self, parser, cli_name):
+        pass
+
+    def main(self, *, args):
+        raise NotImplementedError()
--- a/README.md
+++ b/README.md
@@ -133,6 +133,18 @@ By default, it will enable all ROS 2 tracepoints.
 The trace will be written to `~/.ros/tracing/session-YYYYMMDDHHMMSS`.
 Run the command with `-h` for more information.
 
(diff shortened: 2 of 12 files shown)


christophebedard:
@mjcarroll friendly ping

ros2trace/ros2trace/verb/__init__.py
```diff
@@ -0,0 +1,44 @@
+# Copyright 2017 Open Source Robotics Foundation, Inc.
```
mjcarroll, line 1: I suppose this is a copy from somewhere else, but still a new file that you authored?

```suggestion
# Copyright 2023 Apex.AI, Inc.
```

mjcarroll (approved):
LGTM, just one copyright nit

ros2trace/ros2trace/verb/__init__.py
christophebedard: Yeah, I copied it from one of the `ros2cli` packages and only changed the name of the command (in single quotes) in the class docstring. I didn't change the copyright owner because of that, but I guess I can change it if you think I should.

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19125/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13660/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19863/)

christophebedard:
Thanks for the review.

CI looks good, so let's go.

ros-discourse:

PR #53, opened 2023-04-05
Disable tracing on macOS

christophebedard:
Fixes #52

Prior to #31, `tracetools` tried to look for LTTng-UST on macOS and Linux, and just disabled tracing if it wasn't found. However, LTTng-UST is now required. Just like we do for Windows, just disable tracing and don't try to look for LTTng on macOS.

Issue #52, opened 2023-04-05
regression in build on macos


wjwwood:
Since https://github.com/ros2/ros2_tracing/pull/31 (I think it's this one) this no longer builds on macos:

```
% colcon build --symlink-install --merge-install --cmake-args -DPython3_EXECUTABLE=`which python3` -DPYTHON_EXECUTABLE=`which python3` --mixin compile-commands debug --event-handlers console_direct+ console_stderr- --packages-select tracetools
Starting >>> tracetools
Not searching for unused variables given on the command line.
-- Found ament_cmake_ros: 0.11.2 (/Users/william/2011_ws/install/share/ament_cmake_ros/cmake)
-- Override CMake install command with custom implementation using symlinks instead of copying resources
-- Checking for module 'lttng-ust'
--   No package 'lttng-ust' found
CMake Error at /opt/homebrew/Cellar/cmake/3.23.2/share/cmake/Modules/FindPkgConfig.cmake:605 (message):
  A required package was not found
Call Stack (most recent call first):
  /opt/homebrew/Cellar/cmake/3.23.2/share/cmake/Modules/FindPkgConfig.cmake:827 (_pkg_check_modules_internal)
  CMakeLists.txt:36 (pkg_check_modules)


-- Configuring incomplete, errors occurred!
See also "/Users/william/2011_ws/build/tracetools/CMakeFiles/CMakeOutput.log".
See also "/Users/william/2011_ws/build/tracetools/CMakeFiles/CMakeError.log".
Failed   <<< tracetools [0.45s, exited with code 1]

Summary: 0 packages finished [1.09s]
  1 package failed: tracetools
  1 package had stderr output: tracetools
```

Reverting to e664e5b let me build cleanly (with some compiler warnings):

```
% colcon build --symlink-install --merge-install --cmake-args -DPython3_EXECUTABLE=`which python3` -DPYTHON_EXECUTABLE=`which python3` --mixin compile-commands debug --event-handlers console_direct+ console_stderr- --packages-select tracetools
Starting >>> tracetools
Not searching for unused variables given on the command line.
-- Found ament_cmake_ros: 0.11.2 (/Users/william/2011_ws/install/share/ament_cmake_ros/cmake)
-- Override CMake install command with custom implementation using symlinks instead of copying resources
-- Checking for module 'lttng-ust'
--   No package 'lttng-ust' found
[...]

PR #71, opened 2023-04-26
Disable tracing on Android

Nizerlak:
Similarly to #52, there was also an issue with building on android (following steps in [this repo](https://github.com/YasuChiba/ros2-android-build))
I've just simply applied same solution as in #53.

### DIFF
--- a/test_tracetools/CMakeLists.txt
+++ b/test_tracetools/CMakeLists.txt
@@ -13,7 +13,7 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 
-if(WIN32 OR APPLE)
+if(WIN32 OR APPLE OR ANDROID)
   set(DISABLED_DEFAULT ON)
 else()
   set(DISABLED_DEFAULT OFF)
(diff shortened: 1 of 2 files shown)


christophebedard:
Thanks for the PR! While I think installing LTTng on Android should be doable (unlike Windows), I think disabling it by default is reasonable.

We'll just have to wait until the Rolling freeze is lifted in a few days before merging. And then I think we could backport this to the new Iron branch as a bug fix.

Nizerlak:
Should I do anything more regarding this PR?

christophebedard:
> Should I do anything more regarding this PR?

No, I'll pick it up once Rolling is unfrozen.

christophebedard:
Ah, one thing, though: the DCO check job is complaining because the commit author and the sign-off don't match:

* Commit author: `Przemysław Dąbrowski <57660090+Nizerlak@users.noreply.github.com>`
* Sign-off: `Przemyslaw Dabrowski <57660090+Nizerlak@users.noreply.github.com>`

Looks like the only difference is the ą/a. Would you mind changing the sign-off to match the commit author? Diacritics should be fine.

christophebedard (approved):

Issue #73, opened 2023-05-10
Proposal: move dds tracepoints to RWM layers


cwecht:
ros2profile needs a few tracepoints which are currently located in the DDS implementations itself, e.g. here in cyclonedds: https://github.com/christophebedard/cyclonedds/blob/instrumentation-lttng-update/src/core/ddsc/src/tracing_lttng.h

The tracepoints in question are:
- dds:create_writer
- dds:create_reader
- dds:write
- dds:read

AFAIK none of these tracepoints made it to the master branches of the respective DDS implementation. However for commercial products like RTI ConnextDDS it is probably almost impossible to introduce these tracepoints in their code base.
Additionally the current approach imposes a massive maintainance burden.

I had a look at rmw_cyclonedds and rmw_fastrtps and found that it should be possible to add the respective tracepoints there as well. I had added tracepoints to rmw_connextdds before, too:
- https://github.com/ros2/rmw_cyclonedds/commit/816c50682c9f7999de38205e9ee49c70f436e146
- https://github.com/ros2/rmw_fastrtps/commit/f2ac3e0ac05a6994b2daedfd4929a2b99924fd51
- https://github.com/ros2/rmw_connextdds/commit/3facbce260d712afab6000c17216b721b694abcc#diff-15424b0249cc41e5d4179bade17b540ec7d1edd8b20f7d60bc614fe7be0316d3

Please note that
1. these changes are not working right now because I didn't bother for the moment to migrate the tracepoint definitions to the respective repositories. These changes are only for showcaseing what's possible as a basis for discussion.
2. The changes in rmw_cyclonedds are a bit tricky as it is not easily possible to get a pointer to a writer/reader without setting a mutex on them. I just used to rmw-subscriber to fill the pointer field. This should suffice anyways.
3. The current implementation in rmw_connextdds introduces a quite massive performance penalty due DDS_DomainParticipant_get_current_time being really slow. It seems though, that it is possible to use other time souces here.

Based on these findings I propose to:
1. move dds:create_writer" and dds:create_reader to tracetools.
2. move "dds:write" and "dds:read" to tracetools as well or, at least the the source_timestamp field to ros2:rmw_publish, as ros2:rmw_publish and dds:write are usually realy close anyways.

[...]

PR #74, opened 2023-05-15
Add timestamp to rmw_publish tracepoint

cwecht:
See https://github.com/ros2/ros2_tracing/issues/73

Corresponding `rmw` implementation PRs:

* https://github.com/ros2/rmw_cyclonedds/pull/454
* https://github.com/ros2/rmw_fastrtps/pull/694
* https://github.com/ros2/rmw_connextdds/pull/120

action-ros-ci-repos-override: https://gist.githubusercontent.com/christophebedard/124d8311bed71f7820a2185ce0a25cc3/raw/49c81e6cacce1a5e621ed3880b205dac9096a3ac/ros2.repos

### DIFF
--- a/tracetools/include/tracetools/tp_call.h
+++ b/tracetools/include/tracetools/tp_call.h
@@ -137,10 +137,14 @@ TRACEPOINT_EVENT(
   TRACEPOINT_PROVIDER,
   rmw_publish,
   TP_ARGS(
-    const void *, message_arg
+    const void *, publisher_handle_arg,
+    const void *, message_arg,
+    int64_t, timestamp_arg
   ),
   TP_FIELDS(
+    ctf_integer_hex(const void *, publisher_handle, publisher_handle_arg)
     ctf_integer_hex(const void *, message, message_arg)
+    ctf_integer(int64_t, timestamp, timestamp_arg)
   )
 )
 
--- a/tracetools/include/tracetools/tracetools.h
+++ b/tracetools/include/tracetools/tracetools.h
@@ -274,7 +274,9 @@ _DECLARE_TRACEPOINT(
  */
 _DECLARE_TRACEPOINT(
   rmw_publish,
-  const void * message)
+  const void * publisher_handle,
+  const void * message,
+  int64_t timestamp)
 
 /// `rmw_subscription_init`
 /**
--- a/tracetools/src/tracetools.c
+++ b/tracetools/src/tracetools.c
@@ -156,9 +156,13 @@ DEFINE_TRACEPOINT(
 DEFINE_TRACEPOINT(
   rmw_publish,
   TRACEPOINT_PARAMS(
-    const void * message),
+    const void * publisher_handle,
+    const void * message,
+    int64_t timestamp),
   TRACEPOINT_ARGS(
-    message))
+    publisher_handle,
+    message,
+    timestamp))
 
 DEFINE_TRACEPOINT(
   rmw_subscription_init,


christophebedard:
Sorry for the very late follow-up here. Good to see that the `rmw_*` maintainers are in favour of this change.

You can remove `rmw_publisher_init` & `rmw_subscription_init` since they are not needed if we remove (the need for) the DDS tracepoints. See https://github.com/ros2/ros2_tracing/issues/73#issuecomment-1546273243. A rebase would also be good.

My only concern is that this breaks API/ABI. Tracepoints are effectively part of the `tracetools` API and ABI because they are exposed as normal functions, since `TRACETOOLS_TRACEPOINT()` macros expand to normal function calls. I think we usually use the [tick-tock strategy](https://docs.ros.org/en/rolling/The-ROS2-Project/Contributing/Developer-Guide.html#deprecation-strategy) for this kind of breaking change. In the past, for the sake of API/ABI stability, we've kept unused tracepoint fields: see [`rclcpp_publish`](https://github.com/ros2/ros2_tracing/blob/0041e07e034538f7fa99cdcbe7c2d2c83dc1d272/tracetools/include/tracetools/tracetools.h#L233). However, given that this is a real improvement and that there is no good, "clean" way to deal with this breaking change, I would personally lean towards just going ahead with this in Rolling, but I want to check with other core maintainers. @clalancette what do you think?

clalancette:
> However, given that this is a real improvement and that there is no good, "clean" way to deal with this breaking change, I would personally lean towards just going ahead with this in Rolling, but I want to check with other core maintainers. @clalancette what do you think?

I have a few different thoughts about this:

1.  It looks like the changes to `rmw_connextdds` haven't been approved yet.  I'd want to make sure we have agreement from @asorbini before going forward there.
2.  The biggest problem with changing the tracepoints like this would be affecting the out-of-core RMWs.  However, from what I can tell the only out-of-core RMW that uses `TRACEPOINT` is https://github.com/christophebedard/rmw_email, so I guess we are OK there :).
3.  We could consider adding a new tracepoint, `rmw_publish_with_timestamp` or whatever.  Then we could switch the in-core RMWs to use that.  But it is probably not worth it.

Overall, I'm OK with this change, as long as all of the changes to the in-core RMWs are ready and merged at the same time.

christophebedard:
Sounds good!

> 2. The biggest problem with changing the tracepoints like this would be affecting the out-of-core RMWs. However, from what I can tell the only out-of-core RMW that uses TRACEPOINT is https://github.com/christophebedard/rmw_email, so I guess we are OK there :).

I'm sure the author of that RMW implementation wouldn't mind :smile:

christophebedard:
> You can remove `rmw_publisher_init` & `rmw_subscription_init` since they are not needed if we remove (the need for) the DDS tracepoints.

If we weren't changing an rmw tracepoint's fields, we probably wouldn't want to remove these other rmw tracepoints, but we are, so removing them doesn't really make a huge difference. However, we could keep them in case we/people want to instrument middleware implementations for other purposes in the future. Then they could still be used to link ROS 2 pubs/subs to the corresponding middleware objects. So I might leave them in for now.

cwecht:
> 
>     1. It looks like the changes to `rmw_connextdds` haven't been approved yet.  I'd want to make sure we have agreement from @asorbini before going forward there.
> 

Please note that rmw_connextdds hadn't introduced the ros2_tracing tracepoints before https://github.com/ros2/rmw_connextdds/pull/120. From my point of view, rmw_connextdds shouldn't really block us here.

christophebedard:
I agree, but it would block us _if_ the proposed design is rejected/doesn't work for `rmw_connextdds`. However, looks like it should be good.

cwecht:
> I agree, but it would block us _if_ the proposed design is rejected/doesn't work for `rmw_connextdds`. However, looks like it should be good.

https://github.com/ros2/rmw_connextdds/pull/120 has now been approved, too. So every "major RMW implementation" has approved the proposed changes now. However, in the meantime this arised: https://github.com/ros2/rmw_cyclonedds/pull/454#discussion_r1428212218 . IMO we should be fine for the time being with the proposed solution, but I'd like to have @christophebedard s approval here.

cwecht:
> > I agree, but it would block us _if_ the proposed design is rejected/doesn't work for `rmw_connextdds`. However, looks like it should be good.
> 
> [ros2/rmw_connextdds#120](https://github.com/ros2/rmw_connextdds/pull/120) has now been approved, too. So every "major RMW implementation" has approved the proposed changes now. However, in the meantime this arised: [ros2/rmw_cyclonedds#454 (comment)](https://github.com/ros2/rmw_cyclonedds/pull/454#discussion_r1428212218) . IMO we should be fine for the time being with the proposed solution, but I'd like to have @christophebedard s approval here.

Quick update on this one: it turned out, that the issue mentioned in https://github.com/ros2/rmw_cyclonedds/pull/454#discussion_r1428212218 could be fixed by using a slightly different API from cyclone: https://github.com/ros2/rmw_cyclonedds/pull/454#discussion_r1439240556 . At this point its probably more on the rmw_cyclonedds maintainers to give their approval regarding this change.

christophebedard:
Looks good from my end! Just waiting on an approval for the last tiny change for the Cyclone DDS PR and then I'll probably run CI again. However, if you could amend and force-push here (without changing anything), it would correctly run local CI with the corresponding `rmw` PRs.

`test_tracetools`' `test_publisher.py` and `test_pub_sub.py` should be updated to cover the new `rmw_publish` fields, but that could be done in another PR (I could even do it), I don't want to delay this PR any more :sweat_smile:

cwecht:
We've got the CycloneDDS PR approved: https://github.com/ros2/rmw_cyclonedds/pull/454#pullrequestreview-1815748043

christophebedard:
Full CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20077/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14526/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20796/)

christophebedard (approved):
ROS 2 CI looks good. I don't have the necessary permissions to merge the rmw PRs, so I'll let @mjcarroll merge everything at once.

fujitatomoya (approved):

christophebedard:
@mjcarroll friendly ping

christophebedard:
Thanks for merging! I'll create a new release now.

Thanks for the contribution @cwecht!

christophebedard:
> I'll create a new release now.

https://github.com/ros/rosdistro/pull/39635

ros-discourse:

PR #81, opened 2023-10-26
Switch to custom lttng-ctl Python bindings

christophebedard:
Requires https://github.com/ros2/ci/pull/725 for CI

This replaces the lttng-ctl Python bindings (`python3-lttng`), on which the `ros2 trace` command and `Trace` launch action rely, with a new implementation.

The new lttng-ctl Python bindings implementation, `lttngpy`, uses `pybind11` to bind C++ code to Python. The bindings do not cover the whole trace control library (`liblttng-ctl`, from the `liblttng-ctl-dev` package on Ubuntu), but they cover whatever is needed by `tracetools_trace`/`ros2trace`. A lot of `lttngpy` is just exposing C functions and constants from `lttng-ctl` (which usually start with `lttng_*` or `LTTNG_*`) to Python. However, `lttngpy` does wrap some `lttng-ctl` functions to provide a more Pythonic interface, e.g., `Optional[...]` arguments and `Union[...]` return types.

I replaced the calls to the current Python bindings (`lttng`) in `tracetools_trace` (and some other locations) with calls to `lttngpy`. It is (almost) a simple drop-in replacement; users should not notice any difference. In fact, this even improves the usefulness of `ros2 trace`/`Trace`, since they now support enabling [perf counters](https://lttng.org/docs/v2.13/#doc-adding-context), e.g., `ros2 trace -c perf:thread:task-clock`. Indeed, the bindings available from `python3-lttng` do not completely cover `liblttng-ctl`. They are always lagging behind; I've fixed issues and contributed improvements before, but, since any new improvements are only going to be available from `apt` in the next Ubuntu LTS release, we cannot rely on this. Therefore, I think implementing our own bindings is a better solution.

Note that RHEL has `liblttng-ctl` version 2.12.11, while Ubuntu has 2.13.x, so I had to adapt some things to support both 2.12 and 2.13.

Note: The _Sanity checks / binary (rolling)_ CI job will fail because `ros-rolling-lttngpy` is of course not yet available.


christophebedard:
@iluetkeb as we discussed

iluetkeb (approved):
Awesome work, thanks :-)

lttngpy/src/lttngpy/channel.cpp
```diff
@@ -0,0 +1,87 @@
...
+  const std::optional<uint64_t> read_timer_interval,
+  const std::optional<enum lttng_event_output> output)
+{
+  struct lttng_domain domain {};
+  domain.type = domain_type;
+  domain.buf_type = buffer_type;
+
+  struct lttng_handle * handle = lttng_create_handle(session_name.c_str(), &domain);
+  if (nullptr == handle) {
+    return -LTTNG_ERR_UNK;
```
mjcarroll, line 48: Just out of curiosity, why negative here?

mjcarroll (approved):
Overall the approach looks good.  While it isn't zero-cost, I think writing your own bindings is the right call in this situation, especially when dealing with LTS durations.

I had one small question of something that jumped out at me as kind of odd, but overall LGTM

lttngpy/src/lttngpy/channel.cpp
christophebedard: That's the behaviour of the `lttng_enable_channel()` function that this wraps: https://github.com/lttng/lttng-tools/blob/d4333eb996dec315fb721074c2cf49adee04363b/include/lttng/channel.h#L87. The documentation for `lttngpy::enable_channel()` says the same thing: https://github.com/ros2/ros2_tracing/blob/ca833252f8e35bf86c0da7ff7e68678ed796b36e/lttngpy/src/lttngpy/channel.hpp#L41

The lttng-ctl error code constants are positive: https://github.com/lttng/lttng-tools/blob/d4333eb996dec315fb721074c2cf49adee04363b/include/lttng/lttng-error.h#L24-L217. When used as a return value, they are negated. A positive return value means that the call was successful, see for example `lttng_list_channels()`: https://github.com/lttng/lttng-tools/blob/d4333eb996dec315fb721074c2cf49adee04363b/include/lttng/channel.h#L77C74-L77C74. `lttng_strerror()` can handle both negative and positive error codes, though.

christophebedard:
Testing everything above `lttngpy`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19740/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14244/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20466/)

christophebedard:
Oh, right, I need to install `liblttng-ctl-dev` in CI.

christophebedard:
@mjcarroll I don't have access to ros2/ci. Could you push my PR branch to the repo? Then I can re-run the CI jobs above using that CI branch, or you can do it if you want.

mjcarroll:
https://github.com/ros2/ci/tree/christophebedard/install-liblttng-ctl-dev

christophebedard:
The RHEL job fails due to another issue, see https://github.com/ros2/ci/pull/725#issuecomment-1819548255.

christophebedard:
Let's try again:

Testing everything above `lttngpy` with https://github.com/ros2/ci/pull/725:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19973/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14445/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20683/)
* Linux RHEL [Build Status](http://ci.ros2.org/job/ci_linux-rhel/545/)

christophebedard:
Looking good now, RHEL included.

@mjcarroll could you review the latest changes, i.e., the last commit?

mjcarroll (approved):

ros-discourse:

PR #83, opened 2023-11-26
Switch to target_link_libraries in test_tracetools.

clalancette:
While we are in here, mark all dependencies as test_depend (since this is a test-only package).

### DIFF
--- a/test_tracetools/CMakeLists.txt
+++ b/test_tracetools/CMakeLists.txt
@@ -42,64 +42,72 @@ if(BUILD_TESTING)
   add_executable(test_publisher
     src/test_publisher.cpp
   )
-  ament_target_dependencies(test_publisher
-    rclcpp
-    std_msgs
+  target_link_libraries(test_publisher PRIVATE
+    rclcpp::rclcpp
+    ${std_msgs_TARGETS}
   )
+
   add_executable(test_intra
     src/test_intra.cpp
   )
-  ament_target_dependencies(test_intra
-    rclcpp
-    std_msgs
+  target_link_libraries(test_intra PRIVATE
+    rclcpp::rclcpp
+    ${std_msgs_TARGETS}
   )
+
   add_executable(test_lifecycle_node
     src/test_lifecycle_node.cpp
   )
-  ament_target_dependencies(test_lifecycle_node
-    rclcpp
-    rclcpp_lifecycle
+  target_link_libraries(test_lifecycle_node PRIVATE
+    rclcpp::rclcpp
+    rclcpp_lifecycle::rclcpp_lifecycle
   )
+
   add_executable(test_lifecycle_client
     src/test_lifecycle_client.cpp
   )
-  ament_target_dependencies(test_lifecycle_client
-    lifecycle_msgs
-    rclcpp
+  target_link_libraries(test_lifecycle_client PRIVATE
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19878/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14364/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20587/)

christophebedard:
It looks like this breaks something in `test_ros2trace`, which relies on `test_tracetools`. I see the same failure in GitHub CI: https://github.com/ros2/ros2_tracing/actions/runs/6998187248/job/19035964490?pr=83#step:4:28701. I'll try to take a look today.

clalancette:
> [`test_ros2trace` uses node executables from `test_tracetools`](https://github.com/ros2/ros2_tracing/blob/0d9883c183244d4ac8f35e0e759b0642f5f91014/test_ros2trace/test/test_ros2trace/test_trace.py#L217-L234), so `test_ros2trace` depended on `test_tracetools` `<[exec_]depend>`ing on `std_msgs`, `rclcpp`, and friends: https://github.com/ros2/ros2_tracing/actions/runs/6998187248/job/19035964490?pr=83#step:4:28395

Thanks for this analysis, it was helpful.

While looking at this, it occurred to me that there was another solution here.  And that is to export the dependencies that we need from `test_tracetools` so that colcon will properly setup the environment.  That way we can keep all of the dependencies as `test_depend`, but just export the ones that are needed externally.  I've done that in 819c592, and it seems to work in my testing.

Thoughts on this approach?

christophebedard:
Sorry for the delay. That's a good idea.

However, if we want `test_tracetools` test node executables to be used by tests in other `test_*` packages, then we should do this for `std_srvs` and `rclcpp_lifecycle` too (I think those are the other required dependencies if a downstream package were to use all nodes). Explicitly mentioning `test_ros2trace` in `test_tracetools` and limiting the `<build_export_depend>` to just those dependencies feels a bit weird, but I don't really mind. We can always change it in the future; it's just a test package.

christophebedard (approved):

clalancette:
> However, if we want `test_tracetools` test node executables to be used by tests in other `test_*` packages, then we should do this for `std_srvs` and `rclcpp_lifecycle` too (I think those are the other required dependencies if a downstream package were to use all nodes). Explicitly mentioning `test_ros2trace` in `test_tracetools` and limiting the `<build_export_depend>` to just those dependencies feels a bit weird (i.e., `test_tracetools` ideally shouldn't know about downstream packages), but I don't really mind. We can always change it in the future; it's just a test package.

Good points.  I've updated it now to `build_export_depend` all of the necessary dependencies, and I've also updated the language to be less specific to test_ros2trace.  I'm going to pull this out of draft status now and run CI on it, thanks for the review!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19909/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14388/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20617/)

christophebedard (approved):
Looks good!

PR #84, opened 2023-12-02
Fix use of mutable default arg in tracetools_test

christophebedard:
This issue only affects `tracetools_test`-based tests if multiple tests are run in the same `pytest` invocation (i.e., `ament_add_pytest_test()`). The issue is that the default `list` value for `additional_actions` is mutable. We kept appending to the same `list`, so we re-used `Trace` action objects for multiple `LaunchService.run()` calls, and ended up performing substitutions twice, which was raising an error.

I fixed this by changing the default value to `None` and then changing `None` to `[]` if needed. I also changed `additional_actions` to be appended instead of prepended to the list of launch actions, since the `Trace` action must be first.

I also simplified the typing a bit. This `additional_actions` argument isn't actually used anywhere here, so it could technically be removed, but we can just keep it.

Note that `test_tracetools` currently has individual `ament_add_pytest_test()` calls for its tests, but I've bumped into this issue internally.


christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19923/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14401/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20633/)

christophebedard:
CI looks good. Thank you for the review.

Issue #27, opened 2022-11-23
Fix mypy test failures


christophebedard:
In `tracetools_launch`:

```
  tracetools_launch/actions/ld_preload.py:23:1: error: Module "launch" has no attribute "logging"  [attr-defined]
  tracetools_launch/action.py:28:1: error: Module "launch" has no attribute "logging"  [attr-defined]
  test/tracetools_launch/test_ld_preload_action.py:21:1: error: Module "launch" has no attribute "LaunchDescription"  [attr-defined]
  test/tracetools_launch/test_ld_preload_action.py:22:1: error: Module "launch" has no attribute "LaunchService"  [attr-defined]
  launch/example.launch.py:17:1: error: Module "launch" has no attribute "LaunchDescription"  [attr-defined]
  Found 5 errors in 4 files (checked 13 source files)
```

and `tracetools_test`:

```
 tracetools_test/case.py: note: In member "assertMatchingField" of class "TraceTestCase":
  tracetools_test/case.py:318:36: error: Incompatible default for argument "matching_event_name" (default has type "None", argument has type "str")  [assignment]
  tracetools_test/case.py:318:36: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
  tracetools_test/case.py:318:36: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
  tracetools_test/case.py:319:35: error: Incompatible default for argument "events" (default has type "None", argument has type "List[Any]")  [assignment]
  tracetools_test/case.py:319:35: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
  tracetools_test/case.py:319:35: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
  tracetools_test/case.py: note: In member "get_events_with_name" of class "TraceTestCase":
  tracetools_test/case.py:410:35: error: Incompatible default for argument "events" (default has type "None", argument has type "List[Any]")  [assignment]
  tracetools_test/case.py:410:35: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
  tracetools_test/case.py:410:35: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
  tracetools_test/case.py: note: In member "get_events_with_procname" of class "TraceTestCase":
[...]

PR #89, opened 2024-01-30
Fix warnings when using mypy 1.8.0.

clalancette:
For reasons I admit I don't understand, when using mypy 1.8.0 with this code it fails to determine that it can import submodules from launch properly.  Since we can easily workaround this by just doing 'import launch', do that here.

### DIFF
--- a/tracetools_launch/launch/example.launch.py
+++ b/tracetools_launch/launch/example.launch.py
@@ -14,13 +14,13 @@
 
 """Example launch file for the Trace action."""
 
-from launch import LaunchDescription
+import launch
 from launch_ros.actions import Node
 from tracetools_launch.action import Trace
 
 
 def generate_launch_description():
-    return LaunchDescription([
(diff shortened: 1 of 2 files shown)


christophebedard (approved):
I think I've come across these errors (or similar ones) before in #27. It might be due to the `[no-]implicit-reexport` flag: https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport. So it might not like `import`s of an implicitly-reexported value even though it works in practice.

Nevertheless, this is fine.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20188/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14608/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20899/)

PR #93, opened 2024-03-05
Make tracing test assert messages more descriptive

christophebedard:
This is in the context of tracing tests failing on ROS 2 CI after parallelized testing was enabled. The goal is to improve the failed assertion messages to better understand how/why the tests failed just by looking at the test logs, which should hopefully help us fix the tests.

To do this, I:

1. Improved some of the default messages in `TraceTestCase`'s `assert*()` methods to include more information, like the full event(s) data, the field name(s), and the value(s) (if applicable).
2. Removed some of the assert messages in the tests to rely on the default ones. Most (if not all) messages didn't contain the actual event data or value, which is really the part that's useful. The messages only explained the context, which we can figure out by looking at the code anyway, so it wasn't really that useful.

I also did some formatting/code style fixes.

### DIFF
--- a/test_tracetools/test/test_buffer.py
+++ b/test_tracetools/test/test_buffer.py
@@ -61,11 +61,7 @@ class TestBuffer(TraceTestCase):
                 target_buffer,
                 buffer_to_ipb_events)
             # Only 1 for our given buffer
-            self.assertNumEventsEqual(
-                target_buffer_to_ipb_event,
-                1,
-                'none or more than 1 buffer_to_ipb event for the buffer',
-            )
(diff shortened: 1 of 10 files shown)


fujitatomoya (approved):
lgtm with green CI

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20422/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14824/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21165/)

PR #95, opened 2024-03-12
Allow tracing tests to be run in parallel with other tests

### LINKED ISSUE #94: Allow tracing tests to be run in parallel with tracing tests & normal tests
Parallel testing was enabled in ci.ros2.org in https://github.com/ros2/ci/pull/723. Since then, many of the tracing-related tests in `ros2_tracing` have been failing. For example, see the failed tests for March 6th here: https://github.com/osrf/buildfarm-tools/issues/31#issuecomment-1980674852.

There are 2 issues:

1. There is a fundamental issue with running tracing tests in parallel with other tests (any ROS 2 app/executable). When one of the tracing tests starts tracing, it collects data from any running application. As they are written, the tests do not expect to collect unrelated data, and will often fail because of it (e.g., unexpected number of trace events). This can probably be improved and/or fundamentally fixed.
2. Some tracing tests do not expect other tests to control the tracer at the same time as them, and will fail because of it. This constraint was originally meant to circumvent issue 1, but it means that two packages containing tracing tests (which there are) cannot run simultaneously. Once issue 1 is fixed, then this is easy to change.

I changed the tracing test code to provide more information when a tracing test fails due to incorrect/unexpected trace data in #93. We can see the result in the March 6th nightlies, which confirms issue 1:

<details><summary>Test log: test_lifecycle_node.TestLifecycleNode</summary>

```
23:01:14 6: =================================== FAILURES ===================================
23:01:14 6: __________________________ TestLifecycleNode.test_all __________________________
23:01:14 6: 
23:01:14 6: self = <test_lifecycle_node.TestLifecycleNode testMethod=test_all>
23:01:14 6: 
23:01:14 6:     def test_all(self):
23:01:14 6:         # Check events as set
23:01:14 6:         self.assertEventsSet(self._events_ros)
23:01:14 6:     
23:01:14 6:         # Check that the lifecycle node has an init event just like a normal node
23:01:14 6:         rcl_node_init_events = self.get_events_with_name(tp.rcl_node_init)
23:01:14 6:         lifecycle_node_matches = self.get_events_with_field_value(
23:01:14 6:             'node_name',
23:01:14 6:             'test_lifecycle_node',
23:01:14 6:             rcl_node_init_events,
23:01:14 6:         )
23:01:14 6: >       self.assertNumEventsEqual(lifecycle_node_matches, 1)
23:01:14 6: 
23:01:14 6: test/test_lifecycle_node.py:49: 
23:01:14 6: _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
23:01:14 6: ../../../../install/tracetools_test/lib/python3.10/site-packages/tracetools_test/case.py:327: in assertNumEventsEqual
23:01:14 6:     self.assertEqual(len(events), expected_number, msg)
23:01:14 6: E   AssertionError: 2 != 1 : unexpected number of events: [{'_name': 'ros2:rcl_node_init', '_timestamp': 1709708441732661082, 'vpid': 319439, 'rmw_handle': 94411761626944, 'vtid': 319439, 'namespace': '/', 'cpu_id': 1, 'node_handle': 94411760046688, 'procname': 'test_lifecycle_', 'node_name': 'test_lifecycle_node'}, {'_name': 'ros2:rcl_node_init', '_timestamp': 1709708447683372035, 'vpid': 321945, 'rmw_handle': 94716619878512, 'vtid': 321945, 'namespace': '/', 'cpu_id': 2, 'node_handle': 94716619735216, 'procname': 'simple_lifecycl', 'node_name': 'test_lifecycle_node'}]
```

</details>

The test expects 1 event but gets 2 events. We can see that the PIDs (`vpid`) are different, and that the procname of the second event is `simple_lifecycl` with a `test_lifecycle_node` node name, which corresponds to this `ros2lifecycle` test: https://github.com/ros2/ros2cli/blob/d0c8c35f7255f826b542f0d9283d9b706cb2b522/ros2lifecycle/test/test_cli.py#L137-L150.

----

We need to let other ROS 2 tests/applications run while the tracing test runs & traces its test application(s), so we need to either only trace the relevant applications or only consider tracing data from the relevant applications. I can think of a few solutions:

1. Be more aggressive about filtering out unrelated trace data in the test code
    - This would most likely require a lot more work/test code and it would be repetitive; doesn't scale well
2. Use LTTng's [recording event rules feature](https://lttng.org/docs/v2.13/#doc-enabling-disabling-events), e.g., `lttng enable-event -k event_name --filter '$ctx.procname == "string"'`
    - We would only be able to filter on the `procname` value, but there could be collisions, since the same test executables are used by many different tests and `procname` is limited to 15 characters
    - While filtering on PID would be ideal, the process ID isn't known before we launch the application, and we need to start tracing before we launch it
    - Also, the recording event rules feature is not exposed through the `lttngpy` Python bindings, so we would need to add support for it
3. Only consider trace events of the test process' child processes
    - This doesn't work because of course the child processes have terminated by the time we read the trace data
    - The only way this would work is if the parent process gets a list of its child processes _at the right time_, when they're still alive; I don't know how fragile this would be
    - And this assumes that the test process directly spawned the traced processes, which may be a fragile assumption
4. Set `procname` of spawned processes through `Node(exec_name='...')` and only consider events with that `procname`
    - Pretty simple, however this is limited to 15 characters, and some tests already expect the `procname` to correspond to the executable name
5. Use an "active" marker in the trace data itself: have the test generate a unique ID and pass it to the test executable processes being launched by the test using an environment variable, then make the test application record that unique ID in the trace using the simple [`lttng_ust_tracef` API](https://lttng.org/docs/v2.13/#doc-tracef), and have the test code filter out trace data from any process that hasn't recorded that unique ID
    - This would be the most robust solution
    - We would need to modify code, but it's only a few lines and is limited to the test applications (all executables in the `test_tracetools` package)
    - All tracing tests use test applications/executables from the `test_tracetools` package, but not all tracing tests are done with `tracetools_test` for `test_tracetools`; I need probably need to make this filtering logic to other tests (mainly `test_ros2trace`?)
6. Any other solution?

This will then allow us to remove the constraint that led to issue 2.

christophebedard:
Resolves #94

This allows tracing tests to be run in parallel with (a) other tracing tests (e.g., from another package) and (b) normal tests (e.g., any ROS 2 tests). It achieves that by doing two things in two separate commits:

1. Only consider test trace data corresponding to the current test
    1. If we configure tracing in a tracing test and other tests (and therefore "ROS 2 applications") run in parallel, we will collect trace data from all running applications and end up with irrelevant data. #94 provides an overview of the possible solutions along with pros and cons. The chosen (and most robust) solution is:
        1. The tracing test passes a trace test ID to the test application processes it spawns through an environment variable
            1. The trace test ID contains the tracing test's PID and the test name (for easy identification)
        2. The test application processes spawned by the test (all in `test_tracetools`) retrieve that trace test ID from the environment and mark their process by triggering an `lttng_ust_tracef:event` event with that trace test ID in its payload, which will link the trace test ID to that process ID, since the trace event includes the PID
            1. `lttng_ust_tracef:event` was chosen because it is simple and doesn't require defining a new tracepoint; it simply collects a string `msg` value, see the [lttng_ust_tracef API](https://lttng.org/docs/v2.13/#doc-tracef)
        3. The tracing test reads those `lttng_ust_tracef:event` events, collects the PIDs of the events matching its test ID, and filters out trace data from all other processes, leaving only relevant trace data
    1. This only concerns `test_tracetools` and `test_ros2trace`
2. Allow having existing tracing sessions during tracing tests
    1. Now that having multiple tests (in this case multiple tracing tests) run in parallel is possible, we don't have to expect to have only one tracing session at any given time
        1. Since we don't check that _no_ tracing session exists before and after each test, I modified the tests to check that they do not leave their tracing sessions behind
    2. This only concerns `test_ros2trace`

### DIFF
--- a/test_tracetools/CMakeLists.txt
+++ b/test_tracetools/CMakeLists.txt
@@ -36,13 +36,36 @@ if(BUILD_TESTING)
   find_package(lifecycle_msgs REQUIRED)
   find_package(rclcpp REQUIRED)
   find_package(rclcpp_lifecycle REQUIRED)
+  find_package(rcpputils REQUIRED)
   find_package(std_msgs REQUIRED)
   find_package(std_srvs REQUIRED)
+
+  # The utility lib is needed even if TRACETOOLS_DISABLED; it's just empty
+  add_library(${PROJECT_NAME}_mark_process src/mark_process.cpp)
+  if(NOT TRACETOOLS_DISABLED)
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(LTTNG REQUIRED lttng-ust)
+    target_link_libraries(${PROJECT_NAME}_mark_process PRIVATE
+      ${LTTNG_LIBRARIES}
+      rcpputils::rcpputils
+    )
+  endif()
+  target_include_directories(${PROJECT_NAME}_mark_process PUBLIC
+    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+  )
+  if(TRACETOOLS_DISABLED)
+    target_compile_definitions(${PROJECT_NAME}_mark_process PRIVATE TRACETOOLS_DISABLED)
+  endif()
+  install(
+    DIRECTORY include/
+    DESTINATION include/${PROJECT_NAME}
+  )
 
   add_executable(test_publisher
     src/test_publisher.cpp
   )
   target_link_libraries(test_publisher PRIVATE
+    ${PROJECT_NAME}_mark_process
     rclcpp::rclcpp
     ${std_msgs_TARGETS}
   )
@@ -51,6 +74,7 @@ if(BUILD_TESTING)
     src/test_intra.cpp
   )
   target_link_libraries(test_intra PRIVATE
+    ${PROJECT_NAME}_mark_process
     rclcpp::rclcpp
     ${std_msgs_TARGETS}
   )
@@ -59,6 +83,7 @@ if(BUILD_TESTING)
     src/test_lifecycle_node.cpp
   )
   target_link_libraries(test_lifecycle_node PRIVATE
+    ${PROJECT_NAME}_mark_process
     rclcpp::rclcpp
     rclcpp_lifecycle::rclcpp_lifecycle
   )
@@ -67,6 +92,7 @@ if(BUILD_TESTING)
     src/test_lifecycle_client.cpp
   )
   target_link_libraries(test_lifecycle_client PRIVATE
+    ${PROJECT_NAME}_mark_process
     ${lifecycle_msgs_TARGETS}
     rclcpp::rclcpp
   )
@@ -75,6 +101,7 @@ if(BUILD_TESTING)
     src/test_ping.cpp
   )
   target_link_libraries(test_ping PRIVATE
+    ${PROJECT_NAME}_mark_process
     rclcpp::rclcpp
     ${std_msgs_TARGETS}
   )
@@ -83,6 +110,7 @@ if(BUILD_TESTING)
     src/test_pong.cpp
   )
   target_link_libraries(test_pong PRIVATE
+    ${PROJECT_NAME}_mark_process
     rclcpp::rclcpp
     ${std_msgs_TARGETS}
   )
@@ -91,6 +119,7 @@ if(BUILD_TESTING)
     src/test_timer.cpp
   )
   target_link_libraries(test_timer PRIVATE
+    ${PROJECT_NAME}_mark_process
     rclcpp::rclcpp
   )
 
@@ -98,6 +127,7 @@ if(BUILD_TESTING)
     src/test_service_ping.cpp
   )
   target_link_libraries(test_service_ping PRIVATE
+    ${PROJECT_NAME}_mark_process
     rclcpp::rclcpp
     ${std_srvs_TARGETS}
   )
@@ -106,6 +136,7 @@ if(BUILD_TESTING)
     src/test_service_pong.cpp
   )
   target_link_libraries(test_service_pong PRIVATE
+    ${PROJECT_NAME}_mark_process
     rclcpp::rclcpp
     ${std_srvs_TARGETS}
   )
--- a/tracetools_test/tracetools_test/case.py
+++ b/tracetools_test/tracetools_test/case.py
@@ -26,12 +26,17 @@ import unittest
 
 from launch import Action
 from tracetools_read import DictEvent
-from tracetools_read import get_event_name
 from tracetools_read import get_event_timestamp
+from tracetools_read import get_events_with_field_value
+from tracetools_read import get_events_with_name
 from tracetools_read import get_field
 from tracetools_read import get_procname
 from tracetools_read.trace import get_trace_events
 
+from .mark_process import get_corresponding_trace_test_events
+from .mark_process import get_trace_test_id
+from .mark_process import TRACE_TEST_ID_ENV_VAR
+from .mark_process import TRACE_TEST_ID_TP_NAME
 from .utils import cleanup_trace
 from .utils import get_event_names
 from .utils import run_and_trace
@@ -69,7 +74,7 @@ class TraceTestCase(unittest.TestCase):
         super().__init__(methodName=args[0])
         self._base_path = base_path
         self._session_name_prefix = session_name_prefix
-        self._events_ros = events_ros
+        self._events_ros = events_ros + [TRACE_TEST_ID_TP_NAME]
         self._events_kernel = events_kernel
         self._package = package
         self._nodes = nodes
@@ -80,7 +85,10 @@ class TraceTestCase(unittest.TestCase):
         # Get timestamp before trace (ns)
         timestamp_before = int(time.time() * 1000000000.0)
 
-        exit_code, full_path = run_and_trace(
+        # Set trace test ID env var for spawned processes
+        trace_test_id = get_trace_test_id(self._session_name_prefix)
+        os.environ[TRACE_TEST_ID_ENV_VAR] = trace_test_id
+        exit_code, self._full_path = run_and_trace(
             self._base_path,
             self._session_name_prefix,
             self._events_ros,
@@ -91,17 +99,22 @@ class TraceTestCase(unittest.TestCase):
             self._additional_actions,
         )
 
-        print(f'TRACE DIRECTORY: {full_path}')
-        self._exit_code = exit_code
-        self._full_path = full_path
+        print(f'TRACE DIRECTORY: {self._full_path}')
 
         # Check that setUp() ran fine
-        self.assertEqual(self._exit_code, 0)
+        self.assertEqual(exit_code, 0)
 
         # Read events once
-        self._events = get_trace_events(self._full_path)
+        events_all = get_trace_events(self._full_path)
+        self.assertGreater(len(events_all), 0, 'no events found in trace')
+        # Only keep events from processes that have the same marker ID
+        self._events = get_corresponding_trace_test_events(events_all, trace_test_id)
+        self.assertGreater(
+            len(self._events),
+            0,
+            f'no matching trace test events found in trace amongst events: {events_all}',
+        )
         self._event_names = get_event_names(self._events)
-        self.assertGreater(len(self._events), 0, 'no events found in trace')
 
         # Check the timestamp of the first event
         self.assertEventAfterTimestamp(self._events[0], timestamp_before)
@@ -458,7 +471,7 @@ class TraceTestCase(unittest.TestCase):
         """
         if events is None:
             events = self._events
-        return [e for e in events if get_event_name(e) == event_name]
+        return get_events_with_name(event_name, events)
 
     def get_events_with_procname(
         self,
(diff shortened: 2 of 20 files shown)


test_tracetools/CMakeLists.txt
```diff
@@ -36,13 +36,36 @@ if(BUILD_TESTING)
   find_package(lifecycle_msgs REQUIRED)
   find_package(rclcpp REQUIRED)
   find_package(rclcpp_lifecycle REQUIRED)
+  find_package(rcpputils REQUIRED)
   find_package(std_msgs REQUIRED)
   find_package(std_srvs REQUIRED)
 
+  # The utility lib is needed even if TRACETOOLS_DISABLED; it's just empty
+  add_library(${PROJECT_NAME} src/mark_process.cpp)
```
clalancette, line 44: This is minor, but you might want to call this library `mark_process` or something like that.  In other words, `libtest_tracetools.so` doesn't tell me much about what it does (and why a test package is installing it), but `libmark_process.so` is a little more descriptive.

But this is only a suggestion, feel free to ignore it if you prefer to keep it as-is.

tracetools_read/tracetools_read/trace.py
```diff
@@ -64,6 +67,38 @@ def get_trace_events(trace_directory: str) -> List[DictEvent]:
...
+    :param event_name: the event name
+    :param events: the events to check
+    :return: the list of events with the given name
+    """
+    return [e for e in events if get_event_name(e) == event_name]
+
+
+def get_events_with_field_value(
+    field_name: str,
+    field_values: Any,
```
clalancette, line 86: It might be worthwhile to restrict this to `Iterable` types; then you could get rid of the `not isinstance(field_values, (list,set))` below.

But it is just a suggestion, feel free to ignore.

tracetools_test/tracetools_test/case.py
```diff
@@ -78,7 +83,10 @@ def setUp(self):
         # Get timestamp before trace (ns)
         timestamp_before = int(time.time() * 1000000000.0)
 
-        exit_code, full_path = run_and_trace(
+        # Set trace test ID env var for spawned processes
+        trace_test_id = get_trace_test_id(self._session_name_prefix)
+        os.environ[TRACE_TEST_ID_ENV_VAR] = trace_test_id
+        self._exit_code, self._full_path = run_and_trace(
```
clalancette, line 89: This is just nitpicking, but it kind of looks like `self._exit_code` could just be a stack variable, rather than a class member.  That is, it never seems to be used outside of `__init__` here.

clalancette (approved):
Overall, this is absolutely fantastic.  Thanks for taking the time to do this.

I've left a few non-blocking nits inline.  I'm still going to approve, and leave it to you to decide whether to respond to them, and then run full CI on this.

test_tracetools/CMakeLists.txt
christophebedard: Yeah, that's fair. I renamed it to `test_tracetools_mark_process`: https://github.com/ros2/ros2_tracing/compare/cdab11e618ce64a089bdd391699da96894536d55..edeaa9454d15e05865b08d625f15e2c7ecbb8dc7.

tracetools_read/tracetools_read/trace.py
christophebedard: The `Any` is meant to cover both a single value and lists/sets of values. I guess I could change callers to only provide a list/set and not an individual value, but this way the test code is a bit simpler (which was my intention).

tracetools_test/tracetools_test/case.py
christophebedard: Good point, I switched to a local variable: https://github.com/ros2/ros2_tracing/compare/cdab11e618ce64a089bdd391699da96894536d55..edeaa9454d15e05865b08d625f15e2c7ecbb8dc7.

tracetools_read/tracetools_read/trace.py
christophebedard: FYI I moved the two util functions here from `tracetools_read/trace.py` to `tracetools_read/__init__.py`: https://github.com/ros2/ros2_tracing/compare/cdab11e618ce64a089bdd391699da96894536d55..edeaa9454d15e05865b08d625f15e2c7ecbb8dc7. The reason is that the `babeltrace` module is not available on some systems (e.g., Windows), so it cannot be (in)directly imported by tests at the usual file/module level. It's imported indirectly in some test methods, since that is never in scope if LTTng is not installed (see `@unittest.skipIf`) and therefore never in scope if `babeltrace` is not available. Since these two functions don't need `babeltrace`, I moved them to `tracetools_read/__init__.py` with the rest of the util functions to simplify downstream imports.

christophebedard:
Full CI, removing `--executor sequential` in the test args to run tests in parallel (I think that's the only thing I need to do just by looking at https://github.com/ros2/ci/pull/757):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20459/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14861/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21202/)

If the above jobs look good, I will run a RHEL job too.

christophebedard:
The ci_linux-aarch64 job passed, but the ci_linux job is failing with this weird LTTng error I've never seen when trying to destroy tracing sessions:

```
PERROR - 19:49:43.719443455 [372535/372535]: recvmsg: Connection reset by peer (in lttcomm_recv_unix_sock() at unix.c:205)
```

It's a direct failure of `recvmsg()`. @clalancette are you aware of any weird things going on with sockets in CI?

clalancette:
> It's a [direct failure of `recvmsg()`](https://github.com/lttng/lttng-tools/blob/395004a13b038fde13364a662d735d16937ea2d1/src/common/unix.c#L205). @clalancette are you aware of any weird things going on with sockets in CI?

Not off-hand, no.  Also, the aarch64 jobs should be exactly equivalent, in the sense that they use the exact same Dockerfile.  Let's see what happens on the rebuild, which was launched on a different building host.  If we still see it, then we can work with the infrastructure team to try some things out and narrow it down.

christophebedard:
Still failing.

I did realize that an `lttngpy` test tries to destroy all tracing sessions (because that's part of its API), which could affect other tests: https://github.com/ros2/ros2_tracing/blob/ed6f43509b85459b1393a7344aa67b154e2030fa/lttngpy/test/test_session.py#L57. Tests of downstream packages shouldn't be running concurrently, as seen in the CI logs, though. It also wouldn't lead to the error we're getting now.

christophebedard:
I re-triggered the aarch64 job and it failed. I'll try to figure out what tests were running in parallel with the failing tests for the Linux jobs, maybe that will give us a clue.

clalancette:
> I re-triggered the aarch64 job and it failed. I'll try to figure out what tests were running in parallel with the failing tests for the Linux jobs, maybe that will give us a clue.

OK, that sounds good.

If you need, one other thing I can do is to run the tests before and after this change back-to-back on the same machine.  That would at least tell us whether it is an environmental problem or not.  But I'll hold off on doing that until you do some more investigation.

christophebedard:
I rebased this after merging #96. Full parallel CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20478/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14880/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21222/)

I don't expect anything different except that `test_tracetools`'s `test_lifecycle_node` test should not fail, but we'll see.

christophebedard:
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/20478/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14880/)

For both jobs, `test_ros2trace` and `test_tracetools_launch` were running in parallel, but tests only failed in the aarch64 job. Could just be that the tests had less time overlap in the x86_64 job.

clalancette:
So....I made a mistake.

Based on a comment from @christophebedard , I was going to add in a change to this PR that made `test_ros2trace` depend on `test_tracetools_launch`.  It is a bit of a hack, but it should, in theory, force the tracing tests to not run in parallel, thus sidestepping our current parallel problems.

However, by accident I actually pushed that to `rolling` without any review.  Normally I would open a revert PR, but since it is already done and it is a simple one-line change, I've decided to instead rebase this one on top.  I'm going to run CI on this again to see if that improves the situation.  Sorry for the bit of a mess.

clalancette:
Parallel CI tests:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20501/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14903/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21252/)

clalancette:
Given that this is much better than before, I'm going to go ahead and merge this one in.  I'm sure the improvement is a combination of this PR plus what I accidentally pushed, but that is good enough for me at the moment.

PR #96, opened 2024-03-13
Fix interference between test_tracetools and ros2lifecycle tests

christophebedard:
Part of #94

Relates to #95, but is still separate

`test_tracetools`'s `test_lifecycle_node` test and [`ros2lifecycle`'s `test_cli` test](https://github.com/ros2/ros2cli/blob/64d216cb8fafef83d046b79ee6294afb06b7c595/ros2lifecycle/test/test_cli.py) can interfere with each other if run in parallel, making one of them -- or both of them -- fail. For example, `test_tracetools` fails in this nightly job here: https://ci.ros2.org/view/nightly/job/nightly_linux_debug/2988/.

This is mainly because both tests use a lifecycle node with the same name and without any specific namespace:

1. `test_tracetools`'s `test_lifecycle_node` test uses [`test_lifecycle_node`](https://github.com/ros2/ros2_tracing/blob/9a807801f3608f937a3725dc6723a9ac417d670c/test_tracetools/src/test_lifecycle_node.cpp)
2. `ros2lifecycle`'s `test_cli` test [uses](https://github.com/ros2/ros2cli/blob/64d216cb8fafef83d046b79ee6294afb06b7c595/ros2lifecycle/test/test_cli.py#L138-L140) `ros2lifecycle_test_fixtures`'s [`simple_lifecycle_node`, which is named `test_lifecycle_node`](https://github.com/ros2/ros2cli/blob/64d216cb8fafef83d046b79ee6294afb06b7c595/ros2lifecycle_test_fixtures/src/simple_lifecycle_node.cpp#L87)

This can lead to two separate failures:

1. Since the lifecycle nodes have the same name, without any specific namespace, the tests can end up triggering transitions for the wrong lifecycle node
    * I think this sometimes results in `test_tracetools`'s `test_lifecycle_node` test timing out because the state machine ends up in a weird state (not sure why the timeouts in `test_lifecycle_node.cpp` don't make it exit before the 60-second timeout, though)
2. The trace that `test_tracetools`'s `test_lifecycle_node` test collects can include transitions from `ros2lifecycle`

The two separate solutions are:

1. Use a namespace (`/test_tracetools`) for the node in `test_tracetools`'s `test_lifecycle_node` test to avoid transition/topic-level interference
2. Make sure to only consider transition trace events for the right state machine instead of assuming that all transitions in the trace come from the expected state machine and _then_ checking if they correspond to the expected state machine
    * This would also be fixed by #95 even though they are different changes, but this change doesn't really hurt

### DIFF
--- a/test_tracetools/test/test_lifecycle_node.py
+++ b/test_tracetools/test/test_lifecycle_node.py
@@ -33,6 +33,7 @@ class TestLifecycleNode(TraceTestCase):
             ],
             package='test_tracetools',
             nodes=['test_lifecycle_node', 'test_lifecycle_client'],
+            namespace='/test_tracetools',
         )
 
     def test_all(self):
@@ -45,6 +46,11 @@ class TestLifecycleNode(TraceTestCase):
             'node_name',
             'test_lifecycle_node',
             rcl_node_init_events,
+        )
+        lifecycle_node_matches = self.get_events_with_field_value(
+            'namespace',
+            '/test_tracetools',
+            lifecycle_node_matches,
         )
         self.assertNumEventsEqual(lifecycle_node_matches, 1)
         lifecycle_node_handle = self.get_field(lifecycle_node_matches[0], 'node_handle')
@@ -69,7 +75,6 @@ class TestLifecycleNode(TraceTestCase):
 
         # This list of states corresponds to the states that the test_lifecycle_node goes through
         # following test_lifecycle_client's requests
-        # We do not expect any other lifecycle node
         expected_lifecycle_states = [
             'unconfigured',
             'configuring',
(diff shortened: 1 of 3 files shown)


clalancette (approved):
Looks good to me.

For CI, I think it might make sense to run CI tests both sequential and parallel, though I'll leave it up to you.

christophebedard:
Yeah I can do that. Tracing tests might fail in parallel without #95, but this particular test should pass.

christophebedard:
Sequential testing of `tracetools_test`/`test_tracetools` and above only:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20473/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14875/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21216/)

Parallel testing of everything:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20474/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14876/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21217/)

christophebedard:
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14876/)

Those test failures are "expected" (see #95) and not related to this PR.

christophebedard:
Alright, CI results are acceptable, I'll merge this and rebase #95.

PR #100, opened 2024-03-26
Use a memcmp for the expected symbol name.

clalancette:
I believe that g++ does not guarantee what a particular symbol name will be.  Thus, in g++ 11.4.0 (what is in Ubuntu 22.04), the symbol name here ended with "#2", while in g++ 13.2.0 (what is in Ubuntu 24.04), the symbol name ends with "#1".  Given that we can't guarantee this, just search for the first part of the name up to the number, which should be good enough for this test.

### DIFF
--- a/test_tracetools/test/test_utils.cpp
+++ b/test_tracetools/test/test_utils.cpp
@@ -84,10 +84,9 @@ TEST(TestUtils, valid_symbol_lambda_capture) {
 
   auto m = [&](int other_num) {return num + other_num;};
   symbol = tracetools::get_symbol(m);
-  EXPECT_STREQ(
-    symbol,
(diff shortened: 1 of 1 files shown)


christophebedard (changes requested):
Makes sense. Could you just add a short comment in the test to explain why we're not comparing the number?

fujitatomoya (approved):
good information, thanks!

christophebedard (approved):
Thanks for adding the comment!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20613/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14966/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21335/)

PR #102, opened 2024-03-27
Rolling release.

marcoag:
Changelog update and version bump. I will follow up with the tag and bloom once this is merged.

NOTE: We might wanna rebase to keep the commits as per the history in this repo.


christophebedard (approved):
Yes, please rebase to keep the 2 separate commits.

PR #104, opened 2024-03-27
Use Ubuntu 24.04 Noble for GitHub CI

christophebedard:
* The `ubuntu-24.04` GitHub actions runner might not be available until August 2024 (https://github.com/actions/runner-images/issues/9691), so just use Docker
* We seem to now need to explicitly set the default shell to bash to get `source` to work (possibly due to running in Docker and not the runner itself)

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -8,7 +8,9 @@ on:
     - cron: "0 5 * * *"
 jobs:
   binary:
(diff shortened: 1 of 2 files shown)


christophebedard:
Rolling is unfrozen now, so this is ready for review.

christophebedard:
@mergifyio backport jazzy

PR #109, opened 2024-04-09
Improve tracetools_test and simplify test_tracetools code

christophebedard:
The main/biggest improvement is the new `get_event_with_field_value_and_assert()` test method, which can simplify this very common bit of code:

```py
my_events = self.get_events_with_field_value('field', value, events)
self.assertNumEventsEqual(my_events, 1)
my_event = my_events[0]
```

into:

```py
event = self.get_event_with_field_value_and_assert('field', value, events)
```

I've therefore updated a lot of test code to use this new assert method, which makes it a lot simpler and smaller.

I also renamed `assertValidArray()` to `assertValidStaticArray()` and added length checking.

### DIFF
--- a/test_tracetools/test/test_buffer.py
+++ b/test_tracetools/test/test_buffer.py
@@ -56,23 +56,22 @@ class TestBuffer(TraceTestCase):
         # Check corresponding events for construct_buffer_event
         for construct_event in construct_buffer_events:
             target_buffer = self.get_field(construct_event, 'buffer')
-            target_buffer_to_ipb_event = self.get_events_with_field_value(
+            # Only 1 for our given buffer
+            target_buffer_to_ipb_event = self.get_event_with_field_value_and_assert(
                 'buffer',
                 target_buffer,
                 buffer_to_ipb_events,
+                allow_multiple=False,
             )
-            # Only 1 for our given buffer
-            self.assertNumEventsEqual(target_buffer_to_ipb_event, 1)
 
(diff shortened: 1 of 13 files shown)


test_tracetools/test/test_buffer.py
```diff
@@ -56,32 +56,31 @@ def test_all(self):
         # Check corresponding events for construct_buffer_event
         for construct_event in construct_buffer_events:
             target_buffer = self.get_field(construct_event, 'buffer')
-            target_buffer_to_ipb_event = self.get_events_with_field_value(
+            # Only 1 for our given buffer
```
fujitatomoya, line 59: Can we delete this comment, since it is obvious with `allow_multiple=False` argument? and else where.

fujitatomoya (approved):
lgtm with green CI.

test_tracetools/test/test_buffer.py
christophebedard: good idea, see 8f97dedc3c9241db443d3e5936817a2aed6c9bf8. I kept some comments that explained _why_ there should be only 1 event, but removed the ones that basically just said that there's only 1 event.

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20864/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15167/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21550/)

christophebedard:
Thanks for the review.

PR #110, opened 2024-04-10
Include <string> in mark_process.cpp

christophebedard:
To include what we use.

### DIFF
--- a/test_tracetools/src/mark_process.cpp
+++ b/test_tracetools/src/mark_process.cpp
@@ -16,8 +16,12 @@
 #include <lttng/tracef.h>
 #include <lttng/ust-version.h>
 
+#include <string>
+
 #include "rcpputils/env.hpp"
 #endif  // TRACETOOLS_DISABLED
+
+#include <string_view>
 
 #include "test_tracetools/mark_process.hpp"
 


christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20819/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15123/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21504/)

test_tracetools/src/mark_process.cpp
```diff
@@ -19,6 +19,8 @@
 #include "rcpputils/env.hpp"
 #endif  // TRACETOOLS_DISABLED
 
+#include <string>
```
clalancette, line 22: While we are doing this, should we also add:

```
#include <string_view>
```

clalancette (approved):
One comment, but approving anyway.

test_tracetools/src/mark_process.cpp
christophebedard: I was just looking into whether `<string>` includes `<string_view>`, because technically `<string>` is only needed when `#ifndef TRACETOOLS_DISABLED`.

I'll just add `<string_view>` here and move `<string>` inside the `#ifndef TRACETOOLS_DISABLED` above.

christophebedard: Done!

christophebedard:
Thanks for the review, merging.

PR #115, opened 2024-04-17
Fixed warnings - strict-prototypes

ahcorde:
There are new warnings on CI related with this compiler option `strict-prototypes`

https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/1860/gcc/new/

### DIFF
--- a/tracetools/include/tracetools/tracetools.h
+++ b/tracetools/include/tracetools/tracetools.h
@@ -46,20 +46,20 @@
 
 // *INDENT-OFF*
 #  define _TRACEPOINT_NOARGS(event_name) \
-  (ros_trace_ ## event_name)()
+  (ros_trace_ ## event_name)(void)
 #  define _TRACEPOINT_ARGS(event_name, ...) \
   (ros_trace_ ## event_name)(__VA_ARGS__)
 #  define _DO_TRACEPOINT_NOARGS(event_name) \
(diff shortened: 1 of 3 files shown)


tracetools/include/tracetools/tracetools.h
```diff
@@ -103,7 +103,7 @@
  * This is the preferred method over calling the underlying function directly.
  */
 #  define TRACETOOLS_TRACEPOINT_ENABLED(event_name) \
-  ros_trace_enabled_ ## event_name()
+  ros_trace_enabled_ ## event_name(void)
```
christophebedard, line 106: This is not going to work (see build failures in GitHub CI), because these macros are used for both declaration and implementation. I'd need to split them up.

christophebedard: If you wanted to, you could split out the other changes (or remove the changes involving macros above from this PR), since they should be fine:

- `ros_trace_compile_status()` in this file
- `tracetools/src/status_tool.c`
- `tracetools/src/tracetools.c`

Then I can look into dealing with the changes to macros separately.

christophebedard:
@ShravanDeva5327 will take over this as part of #178 and submit a new PR.

christophebedard:
Superseded by #179

PR #116, opened 2024-04-19
Run relevant test_tracetools tests with all instrumented rmw impls

christophebedard:
The `test_tracetools` tracing tests only currently run using the default `rmw` implementation. However, all "main" `rmw` implementations are instrumented, so we should run the tests that cover `rmw` instrumentation against all supported implementations.

Requires #104 for CI to pass.

### DIFF
--- a/test_tracetools/CMakeLists.txt
+++ b/test_tracetools/CMakeLists.txt
@@ -198,6 +198,9 @@ if(BUILD_TESTING)
   # Only run tracing tests if instrumentation and tracepoints are included
   if(NOT TRACETOOLS_TRACEPOINTS_EXCLUDED)
     find_package(ament_cmake_pytest REQUIRED)
+    find_package(rmw_implementation_cmake REQUIRED)
+
+    # Tests to run with the default rmw implementation, which should not matter
     set(_test_tracetools_pytest_tests
       test/test_buffer.py
       test/test_executor.py
@@ -205,12 +208,7 @@ if(BUILD_TESTING)
       test/test_intra_pub_sub.py
       test/test_lifecycle_node.py
       test/test_node.py
-      test/test_pub_sub.py
-      test/test_generic_pub_sub.py
-      test/test_publisher.py
       test/test_service.py
-      test/test_subscription.py
-      test/test_generic_subscription.py
       test/test_timer.py
     )
     foreach(_test_path ${_test_tracetools_pytest_tests})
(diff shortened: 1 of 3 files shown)


test_tracetools/CMakeLists.txt
```diff
@@ -198,19 +198,16 @@ if(BUILD_TESTING)
...
       test/test_buffer.py
       test/test_executor.py
       test/test_intra.py
       test/test_intra_pub_sub.py
       test/test_lifecycle_node.py
       test/test_node.py
-      test/test_pub_sub.py
-      test/test_generic_pub_sub.py
-      test/test_publisher.py
       test/test_service.py
```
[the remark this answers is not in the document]
christophebedard: Services are not yet fully supported/instrumented. If you look at `test_service.py`, you can see that there are no `rmw_*` tracepoints for services.

christophebedard:
Huh, I guess the non-default `rmw` implementations are not available for the Rpr job: https://build.ros2.org/job/Rpr__ros2_tracing__ubuntu_noble_amd64/51/cmake/new/

So if I want the Rpr to stay green, I can't do `message(WARNING ...)`.

christophebedard:
> So if I want the Rpr to stay green, I can't do `message(WARNING ...)`.

I switched to just `message(...)` and now the Rpr job is green.

@fujitatomoya can you take another look?

test_tracetools/package.xml
```diff
@@ -40,6 +40,7 @@
   <test_depend>rclcpp</test_depend>
   <test_depend>rclcpp_lifecycle</test_depend>
   <test_depend>rcpputils</test_depend>
+  <test_depend>rmw_implementation_cmake</test_depend>
```
clalancette, line 43: I think we need a `find_package(rmw_implementation_cmake REQUIRED)` somewhere in the CMakeLists.txt to get access to `get_available_rmw_implementations`.  I'm guessing it works now because it is inherited via something else, but we may as well be explicit about it.

test_tracetools/CMakeLists.txt
```diff
@@ -221,6 +218,37 @@ if(BUILD_TESTING)
...
+      test/test_generic_subscription.py
+      test/test_pub_sub.py
+      test/test_publisher.py
+      test/test_subscription.py
+    )
+    set(_test_tracetools_rmw_implementations
+      rmw_connextdds
+      rmw_cyclonedds_cpp
+      rmw_fastrtps_cpp
+    )
```
clalancette, line 234: Is there a particular reason to choose these, rather than just take whatever `get_available_rmw_implementations` returns?  I don't care a huge amount right now, but as we add in more RMWs, we'll have to make additional changes here which would be nice to avoid.

clalancette (changes requested):
I've left a couple of thoughts inline.

Besides that, do you have an idea of the impact on testing time?  We are going from 5 tests to 15 here, so I just want to make sure this isn't adding a ton of time.

test_tracetools/package.xml
christophebedard: Good catch, I added it.

test_tracetools/CMakeLists.txt
christophebedard: There's a comment above this section of the code that explains it:

> Tests to run with all instrumented/supported rmw implementations

Not all `rmw` implementations are instrumented. For example, `rmw_fastrtps_cpp` is , but `rmw_fastrtps_dynamic_cpp` isn't. Ideally, all core `rmw` implementations would be supported. So once that's done, we could remove this list. I think `rmw_fastrtps_dynamic_cpp` might be the only unsupported core `rmw` implementation.

We could switch this to a disallow-list. However, I believe `get_available_rmw_implementations()` returns all the `rmw` implementations it can find, so if someone has a tier 2+ implementation in their workspace and runs `test_tracetools` tests, then they will probably fail.

Let me know what you think. Maybe we don't need to be this cautious.

christophebedard:
> Besides that, do you have an idea of the impact on testing time? We are going from 5 tests to 15 here, so I just want to make sure this isn't adding a ton of time.

Local GitHub CI (which runs tests sequentially IIRC) only has Cyclone DDS and Fast DDS, so 2 out of 3. For the whole `test_tracetools` package:

1. Before: [42.6 seconds](https://github.com/ros2/ros2_tracing/actions/runs/9181238304/job/25247503987#step:5:27070)
2. After: [55.2 seconds](https://github.com/ros2/ros2_tracing/actions/runs/9508261738/job/26209241137?pr=116#step:5:27266)

If I extrapolate to 2x the time difference (since this went from testing against 1 `rmw` implementation to 2 implementations), we'd get ~67.8 seconds with a third `rmw` implementation. Only 5 of the 13 tracing instrumentation tests get run against the list of 3 `rmw` implementations.

christophebedard:
@ros-pull-request-builder retest this please

christophebedard:
@ros-pull-request-builder retest this please

test_tracetools/CMakeLists.txt
clalancette: All right, that's fine.  We can leave this for now, I was just wondering if we could get rid of this code.

clalancette (approved):
Looks good to me with green CI.

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21298/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15617/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/799/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22032/)

PR #117, opened 2024-05-19
tp_call.h fix field type for tracepoint event rclcpp_ring_buffer_enqueue

mat-kie:
# Reason for this pull request:

While trying to compile ros2 on my system with gcc-14 i got an integer conversion warning causing a compilation failure:

```bash
error: initialization of ‘const uint64_t *’ {aka ‘const long unsigned int *’} from ‘uint64_t’ {aka ‘long unsigned int’} makes pointer from integer without a cast [-Wint-conversion]
  474 |     ctf_integer(const uint64_t *, index, index_arg)
```
# Proposed fix:

As far as i see it, the field type should match the argument type in this case.

After applying the proposed fix locally, the compilation succeeded.

### DIFF
--- a/tracetools/include/tracetools/tp_call.h
+++ b/tracetools/include/tracetools/tp_call.h
@@ -477,7 +477,7 @@ TRACEPOINT_EVENT(
   ),
   TP_FIELDS(
     ctf_integer_hex(const void *, buffer, buffer_arg)
-    ctf_integer(const uint64_t *, index, index_arg)
+    ctf_integer(const uint64_t, index, index_arg)
     ctf_integer(const uint64_t, size, size_arg)
     ctf_integer(const int, overwritten, (overwritten_arg ? 1 : 0))
   )


fujitatomoya (approved):
lgtm.

i guess this is just a bug, https://github.com/ros2/rclcpp/blob/343b29b617b163ad72b9fe3f6441dd4ed3d3af09/rclcpp/include/rclcpp/experimental/buffers/ring_buffer_implementation.hpp#L299

CC: @christophebedard

christophebedard (approved):
Yep, that shouldn't be a pointer. Thanks for the PR!

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21122/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15422/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21839/)

christophebedard:
The failures on aarch64 are due to Mimick and are expected, so this is good. Thanks for the fix!

christophebedard:
@Mergifyio backport iron

PR #120, opened 2024-06-14
Update tp_call.h to fix simple type without cast error

DragonflyRobotics:
While building ROS2 Iron on Arch with an updated PKGBUILD, there seems to be a type mismatch without cast from uin64_t to const uint64_t *, so I just changed it. Perhaps this could be something on my end but I would like to propose the change regardless. Please let me know if you want more details. I am building using colcon with the following PKGBUILD: https://github.com/DragonflyRobotics/ros2-iron-fork. 

Disclaimer: I am not at all familiar with this codebase, but I am familiar with C++. This resolves the compile error but may break functionality, which I cannot anticipate, so any input is appreciated.

### DIFF
--- a/tracetools/include/tracetools/tp_call.h
+++ b/tracetools/include/tracetools/tp_call.h
@@ -471,7 +471,7 @@ TRACEPOINT_EVENT(
   ),
   TP_FIELDS(
     ctf_integer_hex(const void *, buffer, buffer_arg)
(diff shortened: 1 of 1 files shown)


christophebedard:
This was fixed in `rolling` in #117 and then backported to `jazzy` in #119, but not backported to `iron`. I just opened a direct backport here: #121. I'll use that one and close this. Thank you for letting us know that it fails on Iron too.

PR #129, opened 2024-08-02
Fix up two different C problems.

clalancette:
The first one was that the visibility_control file was marked as a C++ header, but was included in a C header. Avoid that by just making visibility_control a C header.

The second problem is that, in C, a function with no arguments technically takes "any" arguments.  Fix
ros_trace_compile_status() to have void arguments, which ensures that no arguments can be passed.

This *might* have an effect on some Clang warnings that we see in e.g. https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/1968/gcc/new/ , but regardless this is more correct.


tracetools/include/tracetools/visibility_control.h
christophebedard, line 1: You might want to change the include guard define from `TRACETOOLS__VISIBILITY_CONTROL_HPP_` to `TRACETOOLS__VISIBILITY_CONTROL_H_`

clalancette: Bah, of course.

christophebedard (approved):
Looks good to me with green CI!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21606/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15894/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1111/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22324/)

fujitatomoya (approved):
lgtm with green CI

PR #137, opened 2024-10-05
Allow enabling syscalls through `ros2 trace` or the Trace action

christophebedard:
This adds support for enabling syscall tracing with the `ros2 trace` command or the `Trace` launch action.

While system calls are separate from kernel events, both can be traced using the LTTng kernel tracer: https://lttng.org/docs/v2.13/#doc-tracing-the-linux-kernel. Kernel events are "tracepoints" while system calls are their own kind of trace event. This explains why this PR adds a separate option for syscalls instead of just using the kernel events option. See the changes to `lttngpy::enable_events` in `lttngpy/src/lttngpy/event.cpp`.

How to use:

1. `ros2 trace` command:
    ```
    $ ros2 trace --syscall openat  # interactive
    $ ros2 trace start my-session-name --syscall openat  # non-interactive
    ```
2. Launch files (Python, XML, and YAML): `syscalls` arg

Note that, when enabling a syscall, e.g., `openat`, the resulting trace has syscall entry and exit events: `syscall_entry_openat` and `syscall_exit_openat`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ $ ros2 trace stop session_name    # Stop tracing after starting or resuming
 
 Run each command with `-h` for more information.
 
-You must [install the kernel tracer](#building) if you want to enable kernel events (using the `-k`/`--kernel-events` option).
+You must [install the kernel tracer](#building) if you want to enable [kernel](https://lttng.org/docs/v2.13/#doc-tracing-the-linux-kernel) events (using the `-k`/`--kernel-events` option) or syscalls (using the `--syscalls` option).
 If you have installed the kernel tracer, use kernel tracing, and still encounter an error here, make sure to [add your user to the `tracing` group](#tracing).
 
 ### Launch file trace action
@@ -185,7 +185,7 @@ $ ros2 launch tracetools_launch example.launch.py
 The `Trace` action will also set the `LD_PRELOAD` environment to preload [LTTng's userspace tracing helper(s)](https://lttng.org/docs/v2.13/#doc-prebuilt-ust-helpers) if the corresponding event(s) are enabled.
 For more information, see [this example launch file](./tracetools_launch/launch/example.launch.py) and the [`Trace` action](./tracetools_launch/tracetools_launch/action.py).
 
-You must [install the kernel tracer](#building) if you want to enable kernel events (`events_kernel` in Python, `events-kernel` in XML or YAML).
+You must [install the kernel tracer](#building) if you want to enable [kernel](https://lttng.org/docs/v2.13/#doc-tracing-the-linux-kernel) events (`events_kernel` in Python, `events-kernel` in XML or YAML) or syscalls (`syscalls` in Python, XML, or YAML).
 If you have installed the kernel tracer, use kernel tracing, and still encounter an error here, make sure to [add your user to the `tracing` group](#tracing).
 
 ## Design
--- a/test_ros2trace/test/test_ros2trace/test_trace.py
+++ b/test_ros2trace/test/test_ros2trace/test_trace.py
@@ -52,6 +52,24 @@ def are_tracepoints_included() -> bool:
     return 0 == process.returncode
 
 
+def skip_if_no_kernel_tracing(func):
+    """Skip test if kernel tracing for kernel tracepoints or syscalls is not available."""
+    def wrapper(*args, **kwargs):
+        kernel_tracepoints = lttngpy.get_tracepoints(domain_type=lttngpy.LTTNG_DOMAIN_KERNEL)
+        syscalls = lttngpy.get_syscalls()
+        error = ', '.join(
+            f'{name}: {lttngpy.lttng_strerror(error_code)}' for name, error_code in (
+                ('kernel tracepoints', kernel_tracepoints),
+                ('syscalls', syscalls),
+            )
+            if isinstance(error_code, int)
+        )
+        if error:
+            raise unittest.SkipTest(f'kernel tracer is required: {error}')
+        return func(*args, **kwargs)
+    return wrapper
+
+
 @unittest.skipIf(not is_lttng_installed(minimum_version='2.9.0'), 'LTTng is required')
 class TestROS2TraceCLI(unittest.TestCase):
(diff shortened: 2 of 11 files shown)


christophebedard:
@mjcarroll would you be willing to review this?

christophebedard:
There's something wrong with the GitHub CI binary jobs: https://github.com/ros2/ros2_tracing/actions/runs/11205771796/job/31145663869?pr=137#step:4:1989. I see the same failures in the nightly GitHub CI binary jobs. Not really related to these changes.

README.md
```diff
@@ -170,7 +170,7 @@ $ ros2 trace stop session_name    # Stop tracing after starting or resuming
 
 Run each command with `-h` for more information.
 
-You must [install the kernel tracer](#building) if you want to enable kernel events (using the `-k`/`--kernel-events` option).
+You must [install the kernel tracer](#building) if you want to enable [kernel](https://lttng.org/docs/v2.13/#doc-tracing-the-linux-kernel) events (using the `-k`/`--kernel-events` option) or syscalls (using the `--syscalls` option).
```
fujitatomoya, line 173: how about using `https://lttng.org/docs/#doc-tracing-the-linux-kernel` here? that is redirected to latest doc version if we do not want to specify the version from this doc.

```suggestion
You must [install the kernel tracer](#building) if you want to enable [kernel](https://lttng.org/docs/#doc-tracing-the-linux-kernel) events (using the `-k`/`--kernel-events` option) or syscalls (using the `--syscalls` option).
```

fujitatomoya (commented):
overall, lgtm with green CI. had a couple of comments.

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -309,6 +316,43 @@ def test_default_tracing(self) -> None:
 
         shutil.rmtree(tmpdir)
 
+    @unittest.skipIf(
+        (
+            isinstance(lttngpy.get_tracepoints(domain_type=lttngpy.LTTNG_DOMAIN_KERNEL), int) or
+            isinstance(lttngpy.get_syscalls(), int)
+        ),
+        'kernel tracer is required',
```
fujitatomoya, line 324: minor comment.

We would want to print the error number here? this only expects `LTTNG_ERR_KERN_NA` but there are many possible cases that `lttng_list_syscalls` fails. that would be more information when something happens to tell us what to do?

README.md
christophebedard: I've thought about this before, because the READE has many links to the LTTng docs. I prefer linking to a specific version in case the docs change in the future. This way, when we switch to a new version of Ubuntu that has a new version of LTTng, I'm forced to update the links and verify that they still point to the right sections. If I don't update the links, then they still point to the right section, but not the right version. Then I think it's easier to find the corresponding section in the docs for the new version, as opposed to finding the section in the new version of the docs from just the `#fragment` (if the link doesn't point to the right section).

test_ros2trace/test/test_ros2trace/test_trace.py
christophebedard: You mean you'd like to include the error codes in the user-facing string?

fujitatomoya: can we use `ttngpy.lttng_strerror` to print the error strings from lltng? i think you already use this several places?

README.md
fujitatomoya: I see that, thanks!

test_ros2trace/test/test_ros2trace/test_trace.py
christophebedard: Absolutely! See the new changes.

Use `--pytest-args -rs` with `colcon test` for it to print the reasons for skipped tests.

fujitatomoya (approved):

christophebedard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21963/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16228/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1477/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22691/)

christophebedard:
Sorry, I realized that the way I had written the decorator meant it was evaluated during test collection, at the very beginning. This means that, if no session daemon was running before the test was run, then the test would not be allowed to try to spawn a session daemon and therefore the test would definitely be skipped. I modified the decorator so that it's evaluated before running the test: https://github.com/ros2/ros2_tracing/compare/485dca3ac871e2a26d603b2916ac3aa6281ebf28..e12b8044bfd1b8a15209df44c10e0f1e47bcf5b0.

Can you take another look?

christophebedard:
Argh this isn't even completely working.

christophebedard:
OK, I fixed the decorator and I think this is all I can really do. Implementing the decorator this way ensures it's evaluated right before running the test, after the test class' `@unittest.skipIf` has been evaluated. Otherwise, running on Windows fails, since `lttngpy` is empty. See the previous Windows CI job: Windows [Build Status](http://ci.ros2.org/job/ci_windows/22691/).

Also, when running with `colcon test` on my system, `test_kernel_tracing` runs, because I have a root session daemon running and my user is part of the `tracing` group. However, if you try to run the test using a build system that does sandboxing, it fails because there's no root session daemon in it, and spawning one would be more tricky. I think this is fine for now.

New CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21965/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16230/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1479/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22693/)

fujitatomoya (approved):
lgtm, thanks for the explanation. lgtm with green CI.

christophebedard:
Thanks for the review @fujitatomoya!

CI looks good.

PR #145, opened 2024-11-17
Instrument client/service for end-to-end request/response tracking

### LINKED ISSUE #143: Instrument service and client for end-to-end request/response tracking
This is similar to tracking messages from publisher to subscription (publication to callback). We want to be able to track a request from the publication/send on the client side to the take & callback on the service side, and similarly for responses.

Some tracepoints already exist:

* `rcl_service_init`
* `rclcpp_service_callback_added`
* `rcl_client_init`

And the following tracepoints are also used for services with `AnyServiceCallback`:

* `rclcpp_callback_register`
* `callback_start`/`callback_end`

The following tracepoints are missing:

* Request send
* Request take
* Response send
* Response take

To track a request/response from end to end, we would ideally use a combination of the client GID + sequence number. With pubs/subs, the publisher GID read from the publisher's side does not match the publisher GID on the subscription's side. However, this may work with clients/services, because the client GID value is sent as-is alongside the request/response; we do not have to rely on the underlying DDS implementation, which could be faulty (https://github.com/ros2/rmw_cyclonedds/issues/377).

Add tests in `test_tracetools`, and update the design document.

christophebedard:
Resolves #143

This adds support for instrumenting clients and services for end-to-end tracking of requests/responses. This means:

1. Associating a request sent by a client to the same request received by a server
2. Associating a response sent by a server to the original request sent by a client
3. Associating a response sent by a server to the same response received by a client
    - There is a slight **limitation** here, see below

The above is achieved with a combination of multiple instrumentation points, mostly in the `rmw` implementations:

1.
    - `rmw_client_init`: On client creation, collect the client's GID
    - `rmw_send_request`: On request publication, collect the client (indirectly corresponding to the client's GID) and the request's sequence number
    - `rmw_take_request`: On request take, collect the request's client GID and sequence number, which can be matched to the published request
1.
    - `rmw_send_response`: On response publication, collect the client GID and the sequence number of the request that the response is for, which can be matched to the published/taken request
1.
    - `rmw_send_response`: On response publication, collect the source timestamp of the response
    - `rmw_take_response`: On response take, collect the client GID and the sequence number of the request that the response is for and the source timestamp of the response, which can be matched to the published/taken request and published response
        - There is a slight **limitation** here, see below

Since clients (and services) are an `rmw` concept and not a DDS concept, selecting the "right" GID to use as the client GID and having it match when read from the client side and the service side was a bit tricky. Refer to the `rmw` implementation PRs (which need to be merged at the same time as this PR):

- https://github.com/ros2/rmw_cyclonedds/pull/521
- https://github.com/ros2/rmw_fastrtps/pull/787
- https://github.com/ros2/rmw_connextdds/pull/163

This updates some tracing tests in `test_tracetools` and adds a new test:

1. `test_client`: basic validation of client tracepoints
2. `test_service`: basic validation of service tracepoints
3. `test_service_req_resp`: validate associations of requests/responses

This also updates the design document.

---

**Limitation**:

 A specific request is, by definition, only sent by a single client and can be received by N>=1 services, but it's still the same request, so there is no issue there. However, there can be multiple responses to the same request, made by multiple services. While having multiple services with the same name is "generally discouraged," it is possible: https://github.com/ros2/rmw/blob/33118c9d4dc2adec838962554f0e09ab5c15d1e0/rmw/include/rmw/rmw.h#L1970-L1972.

To differentiate two service responses to the same request, and therefore know which response the client ends up using, we collect the response's source timestamp when the response is sent and when it is taken. Depending on the clock resolution, we could have collisions here if two services send a response for the same request at the exact same time, but it's unlikely. This is similar to what we do to match a message being published to the same message being received (see #74 and related PRs/issues). We would ideally instead collect the service GID + response sequence number on both sides. However, this is not available for service responses, and DDS-based GIDs are known to be problematic for some `rmw` implementations: https://github.com/ros2/rmw_cyclonedds/issues/377. This is why we simply use source timestamps for pub/sub messages.

---

action-ros-ci-repos-override: https://gist.githubusercontent.com/christophebedard/cf4d163feeeabd2cd38cf694413127a0/raw/1fdc769133ff4c412b8283eebcca729f944f755c/ros2.repos


christophebedard:
> All of this to say: would it be worth also collecting source timestamps of sent/taken responses to differentiate multiple responses to the same request?

I thought about it more and asked some people. I think that, even if it's "discouraged" or rare, because it _can_ happen, it would be good to be able to differentiate multiple responses to the same request.

christophebedard:
I added the source timestamp to the response send/take instrumentation.

This is now ready for review.

christophebedard:
If anyone wants to see how this instrumentation can be used, see this PR that adds support for this new instrumentation in Eclipse Trace Compass (see the included screenshot): https://github.com/eclipse-tracecompass-incubator/org.eclipse.tracecompass.incubator/pull/127.

christophebedard:
@mjcarroll would you be willing to review this?

christophebedard:
Or maybe @clalancette or @fujitatomoya?

fujitatomoya:
@christophebedard i think i can give it a shot.

test_tracetools/test/test_client.py
```diff
@@ -39,11 +42,25 @@ def test_all(self):
         self.assertEventsSet(self._events_ros)
 
         # Check fields
+        rmw_client_init_events = self.get_events_with_name(tp.rmw_client_init)
         client_init_events = self.get_events_with_name(tp.rcl_client_init)
```
fujitatomoya: just a minor suggestion. how about adding `rcl` prefix here too? at least for me it would be easier to understand the order, and other places?

```suggestion
        rcl_client_init_events = self.get_events_with_name(tp.rcl_client_init)
```

tracetools/include/tracetools/tp_call.h
```diff
@@ -299,6 +350,40 @@ TRACEPOINT_EVENT(
   )
 )
 
+TRACEPOINT_EVENT(
+  TRACEPOINT_PROVIDER,
+  rmw_send_request,
+  TP_ARGS(
+    const void *, rmw_client_handle_arg,
```
fujitatomoya, line 357: it would be easier to trace request/response communication if we have `client gid` in here too? (the same comment goes to `rmw_take_response` trace point.)

tracetools/include/tracetools/tracetools.h
```diff
@@ -404,6 +462,46 @@ _DECLARE_TRACEPOINT(
...
+  const void * request,
+  int64_t sequence_number)
+
+/// `rmw_take_response`
+/**
+ * Response taking.
+ * Links a `rmw_client_t` handle to a pointer to a response being taken at the `rmw` level.
+ * Notes the source timestamp of the response. Also notes the sequence number of the request this
+ * response is for. It does not note the request's client GID, since it is assumed that the matching
+ * of the response to the original client is performed before this tracepoint.
```
fujitatomoya, line 487: i believe that it would be easier if we have client gid in here too. i would do grep the client gid to see the whole communication, if it does not show up, it would be kind of confusing?

fujitatomoya:
@christophebedard overall looks good to me, i got several comments though.

test_tracetools/test/test_client.py
christophebedard: that's totally fair; fixed!

tracetools/include/tracetools/tracetools.h
christophebedard: same answer as `rmw_send_request`. Due to how the instrumentation works, you're expected to get the mapping from `rmw` handle to client GID from the initialization tracepoint (`rmw_client_init`). Then following the whole communication from the raw trace data is easy enough, although this data is really meant to be processed.

Also, in general, because of the ROS 2 abstraction layers, we can never get just _one_ single "identifier" that you could use/grep to easily see the path of a message/request/response.

tracetools/include/tracetools/tp_call.h
christophebedard: it would be easier, but it would also be redundant. We can get the client GID from the `rmw_client_handle` thanks to the `rmw_client_init` tracepoint, which provides the mapping from `rmw_client_handle` to (client) `gid`. This is how the instrumentation works in general. This way, at runtime, we can (generally) minimize the size of the data we collect and simplify the tracepoint arguments, since we already have the `rmw_client_handle` and would need to fetch/copy the GID if we used that.

Does that make sense?

fujitatomoya: yeah, it does. i think if it shows the client gid and send request / receive response information close together, that should be no problem. (i was thinking that this could be more consistent and would be easier to search for user if that prints the each message, but efficiency matters as you mentioned.) resolving, thanks for the explanation.

fujitatomoya (approved):
lgtm with green CI.

christophebedard:
Pulls: ros2/ros2_tracing#145, ros2/rmw_cyclonedds#521, ros2/rmw_fastrtps#787, ros2/rmw_connextdds#163
Gist: https://gist.githubusercontent.com/christophebedard/6de7664f9c4b0103faca7eff9ae7a757/raw/f3665ac1140f461336f4c2444f38bacfcc8f44d8/ros2.repos
BUILD args:  --packages-above-and-dependencies test_tracetools tracetools tracetools_trace rmw_cyclonedds_cpp rmw_fastrtps_cpp rmw_fastrtps_dynamic_cpp rmw_fastrtps_shared_cpp rmw_connextdds_common
TEST args:  --packages-above test_tracetools tracetools tracetools_trace rmw_cyclonedds_cpp rmw_fastrtps_cpp rmw_fastrtps_dynamic_cpp rmw_fastrtps_shared_cpp rmw_connextdds_common
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14947
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22259/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16519/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1781/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23038/)

christophebedard:
I had to manually trigger `ci_launcher` because `CI_USE_FASTRTPS_DYNAMIC` needs to be enabled, and Connext DDS is not available on ci_linux-aarch64:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22267/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16528/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1791/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23053/)

christophebedard:
Alright, CI looks good and all PRs are approved. Merging all PRs.

ros-discourse:

https://discourse.ros.org/t/ros-2-visualization-tools-for-architecture/41170/2

PR #147, opened 2024-11-25
Changelog for 8.4.1.

clalancette:
Once this is approved, I will squash-merge it, tag the resulting commit on the `rolling` branch, and do a `bloom-release` of `ros2_tracing`.


christophebedard (approved):

christophebedard:
Since changelog/release PRs are only made on a very small set of core repos, including this one and rosbag2, I'm wondering:

Is the goal of making this PR to get the changelog itself approved, or is it to check (I assume specifically with me) that creating a new release (and deciding whether to bump the minor number or the patch number) is OK, in case it might conflict with my plans? Or maybe a different reason?

clalancette:
> Is the goal of making this PR to get the changelog itself approved, or is it to check (I assume specifically with me) that creating a new release (and deciding whether to bump the minor number or the patch number) is OK, in case it might conflict with my plans? Or maybe a different reason?

It is more the latter; to check with you that this is an OK release to make.

To be perfectly frank, now that we are under one umbrella under the OSRA, it might make sense to get rid of this distinction for `ros2_tracing` and `rosbag2` and just push directly to `rolling` like we do for the other repositories.  It definitely makes life easier when doing "mass" releases into Rolling.

I'll add an item for the ROS PMC meeting tomorrow.

christophebedard:
Agreed, I was asking because I don't think this special treatment is really necessary.

If someone (e.g., ROS boss) is planning to do anything weird or special, release- or changelog-wise, I'd expect them to ping the package's maintainer anyway.

PR #159, opened 2025-03-18
Use ament_add_ros_isolated_* in test_tracetools

christophebedard:
Part of the move towards using `ament_add_ros_isolated_*()` for tests in preparation for `rmw_zenoh` tests.

These are the only tests that explicitly set `RMW_IMPLEMENTATION`.

PR #163, opened 2025-03-24
Set envars to run tests with rmw_zenoh_cpp with multicast discovery

ahcorde:
See https://github.com/ros2/rmw_zenoh/issues/567

To test this PR, we can run the usual CI jobs to show there are no regressions and additionally run a CI job with a copy of the ros2.repos file from https://github.com/ros2/ros2/pull/1649 with CI Scripts branch set to yadu/cargo

### DIFF
--- a/test_tracetools/CMakeLists.txt
+++ b/test_tracetools/CMakeLists.txt
@@ -235,14 +235,28 @@ if(BUILD_TESTING)
       rmw_cyclonedds_cpp
       rmw_fastrtps_cpp
       rmw_fastrtps_dynamic_cpp
+      rmw_zenoh_cpp
     )
     get_available_rmw_implementations(rmw_implementations)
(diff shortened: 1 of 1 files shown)


christophebedard:
This depends on #159 for the `rmw_zenoh_cpp` tests to actually work, right?

Yadunund:
> This depends on #159 for the `rmw_zenoh_cpp` tests to actually work, right?

It technically doesn't since we're overriding rmw_zenoh to run with multicast discovery here. 

@ahcorde I'd like to hold off on the change here to minimize the number of temporary PRs we need to get into the stack.

christophebedard:
We'll go with the long-term solution of using test isolation (#159 + https://github.com/ros2/ament_cmake_ros/pull/28), so I'll close this.

Issue #160, opened 2025-03-19
GitHub CI reports mypy issues


christophebedard:
Some mypy issues started being reported by GitHub CI a few weeks ago in `tracetools_test`, `test_tracetools_launch`, and `tracetools_launch`:

```
tracetools_test/utils.py: note: In function "run_and_trace":
tracetools_test/utils.py:80:31: error: Argument 1 to "append" of "list" has incompatible type "Node"; expected "Trace"  [arg-type]
tracetools_test/utils.py:81:27: error: Argument 1 to "extend" of "list" has incompatible type "list[Action]"; expected "Iterable[Trace]"  [arg-type]

test/test_tracetools_launch/test_trace_action.py: note: In member "_assert_launch_frontend_no_errors" of class "TestTraceAction":
test/test_tracetools_launch/test_trace_action.py:73:16: error: Incompatible return value type (got "LaunchDescriptionEntity", expected "Trace")  [return-value]
test/test_tracetools_launch/test_trace_action.py: note: In member "test_action_context_per_domain" of class "TestTraceAction":
test/test_tracetools_launch/test_trace_action.py:216:13: error: Argument 1 to "assertDictEqual" of "TestCase" has incompatible type "Mapping[str, Iterable[list[Substitution]]] | Iterable[list[Substitution]]"; expected "Mapping[Any, object]"  [arg-type]
test/test_tracetools_launch/test_trace_action.py: note: In member "test_action_substitutions" of class "TestTraceAction":
test/test_tracetools_launch/test_trace_action.py:261:13: error: Argument 1 to "assertDictEqual" of "TestCase" has incompatible type "Mapping[str, Iterable[list[Substitution]]] | Iterable[list[Substitution]]"; expected "Mapping[Any, object]"  [arg-type]
test/test_tracetools_launch/test_trace_action.py: note: In member "test_append_timestamp" of class "TestTraceAction":
test/test_tracetools_launch/test_trace_action.py:345:25: error: "list[Substitution]" has no attribute "startswith"  [attr-defined]

tracetools_launch/actions/ld_preload.py: note: In member "execute" of class "LdPreload":
tracetools_launch/actions/ld_preload.py:84:5: error: Return type "list[Action] | None" of "execute" incompatible with return type "list[LaunchDescriptionEntity] | None" in supertype "Action"  [override]
tracetools_launch/actions/ld_preload.py:86:21: error: List item 0 has incompatible type "Action | None"; expected "Action"  [list-item]
[...]

Issue #149, opened 2024-12-20
Allow creating a live tracing session with `ros2 trace`/`Trace`


christophebedard:
By "live tracing," we mean processing or analyzing the trace data live as it gets generated. This is different from the usual workflow, which is to collect the trace data and then process it offline later, after the execution.

When using the `lttng` CLI, this is done with the `--live` option: https://lttng.org/docs/v2.13/#doc-lttng-live.

It would be good to expose it through the `ros2 trace` command and the `Trace` launch action. A live session can be created using [`lttng-ctl`'s `lttng_create_session_live()`](https://github.com/lttng/lttng-tools/blob/c259edc03bedb9e0d58434cfcc27a7817772507a/include/lttng/session.h#L117-L131); this would need to be exposed through `lttngpy`.

---

In the meantime, here's an example script to enable tracing for ROS 2 with a live session using the `lttng` CLI directly:

<details>

<summary>Click to expand: live_trace.sh</summary>

```sh
# live_trace.sh

if [ $# -eq 0 ]; then
  echo "Provide trace session name"
  exit 1
fi

session_name="$1"
hostname=$(hostname)

lttng create $session_name --live
lttng enable-channel -u ros --subbuf-size=8M
lttng enable-event -c ros -u 'ros2:*'
lttng add-context -u -t vtid -t vpid -t procname
lttng start

echo
echo "Live trace URI: net://localhost/host/$hostname/$session_name"
echo "Press any key to stop & destroy"
read
lttng stop
lttng destroy
```

</details>

Here's a simple Python script that can receive and process some trace data live using the [babeltrace2 Python bindings](https://babeltrace.org/docs/v2.0/python/bt2/index.html) (just an example script):

<details>

<summary>Click to expand: bt2_lttng_live.py</summary>

```py
# bt2_lttng_live.py

import datetime
import sys
from typing import List

import bt2


def handle_trace_event(msg: bt2._EventMessageConst) -> None:
    event = msg.event
    ns_from_origin = msg.default_clock_snapshot.ns_from_origin
    timestamp = datetime.datetime.fromtimestamp(ns_from_origin / 1e9)
    print(f'{timestamp} {event.name}')
    # Accces field with event['field_name']
    # See:
    #   https://babeltrace.org/docs/v2.0/python/bt2/examples.html#get-a-specific-event-field-s-value
    # TODO
    if event.name == 'ros2:rclcpp_intra_publish':
        publisher_handle = event['publisher_handle']
[...]

PR #165, opened 2025-04-05
Some features added to live tracing session.

suchetanrs:
Hello! @christophebedard 

The changes in this PR are as follows:

- Add test to `lttngpy` for creating and destroying a live session
- I changed the `--live` argument and added the arguments `--live-timer-interval` and `--live-tracing-url-origin`
- Live tracing can be triggered with the `start` verb.

### DIFF
--- a/tracetools_trace/tracetools_trace/tools/args.py
+++ b/tracetools_trace/tracetools_trace/tools/args.py
@@ -80,21 +80,14 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         '-a', '--append-trace', dest='append_trace', action='store_true',
         help='append to trace if it already exists, otherwise error out (default: %(default)s)')
-    # parser.add_argument(
-    #     '--live', dest='live_timer_interval', type=int, nargs='?',
-    #     # Default value for 'lttng create-session --live':
-    #     # https://lttng.org/man/1/lttng-create/v2.13/#doc-opt--live
-    #     default=1000000,
-    #     help='TODO (default: %(default)s)')
     parser.add_argument(
-        '--live', dest='live_mode', action='store_true',
-        help='Enable live mode (default: %(default)s)')
-    parser.add_argument(
-        '--live-timer-interval', dest='live_timer_interval', type=int,
-        default=100000,
+        '--live', dest='live_timer_interval', type=int, nargs='?',
+        # Default value for 'lttng create-session --live':
+        # https://lttng.org/man/1/lttng-create/v2.13/#doc-opt--live
+        const=100000,
         help='Set the live timer interval (default: %(default)s)')
     parser.add_argument(
-        '--live-tracing-url-origin', dest='live_tracing_url_origin', type=str,
+        '--live-url', dest='live_url', type=str,
         default='net://localhost',
         help='Set the live tracing URL origin (default: %(default)s)')
 
--- a/tracetools_trace/tracetools_trace/tools/args.py
+++ b/tracetools_trace/tracetools_trace/tools/args.py
@@ -95,8 +95,9 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
         help='Set the live timer interval (default: %(default)s)')
     parser.add_argument(
         '--live-tracing-url-origin', dest='live_tracing_url_origin', type=str,
-        default="net://localhost",
+        default='net://localhost',
         help='Set the live tracing URL origin (default: %(default)s)')
+
 
 def _add_arguments_default_session_name(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
--- a/tracetools_trace/tracetools_trace/tools/args.py
+++ b/tracetools_trace/tracetools_trace/tools/args.py
@@ -85,7 +85,8 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
         # Default value for 'lttng create-session --live':
         # https://lttng.org/man/1/lttng-create/v2.13/#doc-opt--live
         const=100000,
-        help='Set the live timer interval (default: %(default)s)')
+        help='Create a live tracing session. Optionally set the live timer interval '
+             '(default: %(default)s)')
     parser.add_argument(
         '--live-url', dest='live_url', type=str,
         default='net://localhost',
--- a/tracetools_trace/tracetools_trace/tools/args.py
+++ b/tracetools_trace/tracetools_trace/tools/args.py
@@ -80,13 +80,23 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         '-a', '--append-trace', dest='append_trace', action='store_true',
         help='append to trace if it already exists, otherwise error out (default: %(default)s)')
+    # parser.add_argument(
+    #     '--live', dest='live_timer_interval', type=int, nargs='?',
+    #     # Default value for 'lttng create-session --live':
+    #     # https://lttng.org/man/1/lttng-create/v2.13/#doc-opt--live
+    #     default=1000000,
+    #     help='TODO (default: %(default)s)')
     parser.add_argument(
-        '--live', dest='live_timer_interval', type=int, nargs='?',
-        # Default value for 'lttng create-session --live':
-        # https://lttng.org/man/1/lttng-create/v2.13/#doc-opt--live
-        default=1000000,
-        help='TODO (default: %(default)s)')
-
+        '--live', dest='live_mode', action='store_true',
+        help='Enable live mode (default: %(default)s)')
+    parser.add_argument(
+        '--live-timer-interval', dest='live_timer_interval', type=int,
+        default=100000,
+        help='Set the live timer interval (default: %(default)s)')
+    parser.add_argument(
+        '--live-tracing-url-origin', dest='live_tracing_url_origin', type=str,
+        default="net://localhost",
+        help='Set the live tracing URL origin (default: %(default)s)')
 
 def _add_arguments_default_session_name(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
--- a/tracetools_trace/tracetools_trace/tools/lttng_impl.py
+++ b/tracetools_trace/tracetools_trace/tools/lttng_impl.py
@@ -193,10 +193,10 @@ def setup(
     :param subbuffer_size_kernel: the size of the subbuffers for kernel events (defaults to 32
         times the usual page size, since there can be way more kernel events than UST events)
     :param live_timer_interval: the time interval at which the data should be flushed from the
-        buffer and sent to the LTTng relay. This is in microseconds.
-        Used only if live_timer_interval is `True`.
-    :param live_url: the URL to which the tracing output will be sent.
-        Used only if live_timer_interval is `True`.
+        buffer and sent to the LTTng relay daemon. This is in microseconds.
+        The created tracing session will be in live mode if this value is not `None`.
+    :param live_url: the URL of the relay daemon to which the tracing output will be sent.
+        Used only if live_timer_interval is not `None`.
     :return: the full path to the trace directory, or `None` if initialization failed
     """
     # Validate parameters
@@ -258,11 +258,7 @@ def setup(
             full_path=full_path,
         )
     else:
-        # TODO(christophebedard): do we need to join the
-        #   base_path with session_name for a live session?
-        #   We need to return a path, so maybe format it like:
-        #   "net://localhost/host/$hostname/$session_name"
-        live_tracing_url =  'net://localhost/host/' + socket.gethostname() + '/' + session_name
+        live_tracing_url = f'{live_url}/host/{socket.gethostname()}/{session_name}'
         full_path = live_tracing_url
         _create_session_live(
             session_name=session_name,
@@ -458,9 +454,6 @@ def _create_session_live(
     """
     result = lttngpy.lttng_create_session_live(
         session_name=session_name,
-        # TODO(christophebedard): figure out what to provide here as the URL
-        #   This depends on how we expect users to use live tracing
-        #   See the documentation for the url param of lttng_create_session_live()
         url=url,
         timer_interval=timer_interval,
     )
--- a/tracetools_trace/tracetools_trace/tools/lttng_impl.py
+++ b/tracetools_trace/tracetools_trace/tools/lttng_impl.py
@@ -18,6 +18,7 @@
 import os
 import shlex
 import subprocess
+import socket
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -157,7 +158,9 @@ def setup(
     channel_name_kernel: str = 'kchan',
     subbuffer_size_ust: int = 8 * 4096,
     subbuffer_size_kernel: int = 32 * 4096,
+    live_mode: Optional[bool] = False,
     live_timer_interval: Optional[int] = None,
+    live_tracing_url_origin: Optional[str] = None,
 ) -> Optional[str]:
     """
     Set up LTTng session, with events and context.
@@ -190,6 +193,10 @@ def setup(
         the usual page size)
     :param subbuffer_size_kernel: the size of the subbuffers for kernel events (defaults to 32
         times the usual page size, since there can be way more kernel events than UST events)
+    :param live_mode: whether to create a live session
+    :param live_timer_interval: the time interval at which the data should be flushed from the
+        buffer and sent to the LTTng relay. This is in microseconds. Used only if live_mode is `True`.
+    :param live_tracing_url_origin: the URL to which the tracing output will be sent. Used only if live_mode is `True`.
     :return: the full path to the trace directory, or `None` if initialization failed
     """
     # Validate parameters
@@ -199,6 +206,7 @@ def setup(
     # TODO(christophebedard): do we need to join the base_path with session_name for a live session?
     #   We need to return a path, so maybe format it like:
     #   "net://localhost/host/$hostname/$session_name"
+    live_tracing_url = live_tracing_url_origin + '/' + 'host/' + socket.gethostname() + '/' + session_name
     full_path = os.path.join(base_path, session_name)
     if os.path.isdir(full_path) and not append_trace:
         raise RuntimeError(
@@ -247,7 +255,7 @@ def setup(
         raise RuntimeError('no events enabled')
 
     # Create session
-    if live_timer_interval is None:
+    if not live_mode:
         # LTTng will create the parent directories if needed
         _create_session(
             session_name=session_name,
@@ -256,7 +264,7 @@ def setup(
     else:
         _create_session_live(
             session_name=session_name,
-            full_path=full_path,
+            live_tracing_url_origin=live_tracing_url_origin,
             timer_interval=live_timer_interval,
         )
 
@@ -342,6 +350,9 @@ def setup(
             context_fields=contexts_dict.get(domain),
         )
 
+    if live_mode:
+        # TODO(suchetanrs) is there a better way to do this?
+        full_path = live_tracing_url
     return full_path
 
 
@@ -440,7 +451,7 @@ def _create_session(
 def _create_session_live(
     *,
     session_name: str,
-    full_path: str,
+    live_tracing_url_origin: str,
     timer_interval: int,
 ) -> None:
     """
@@ -451,7 +462,7 @@ def _create_session_live(
         # TODO(christophebedard): figure out what to provide here as the URL
         #   This depends on how we expect users to use live tracing
         #   See the documentation for the url param of lttng_create_session_live()
-        url=None,
+        url=live_tracing_url_origin,
         timer_interval=timer_interval,
     )
     if result < 0:
--- a/tracetools_trace/tracetools_trace/tools/lttng_impl.py
+++ b/tracetools_trace/tracetools_trace/tools/lttng_impl.py
@@ -158,9 +158,8 @@ def setup(
     channel_name_kernel: str = 'kchan',
     subbuffer_size_ust: int = 8 * 4096,
     subbuffer_size_kernel: int = 32 * 4096,
-    live_mode: Optional[bool] = False,
     live_timer_interval: Optional[int] = None,
-    live_tracing_url_origin: Optional[str] = None,
+    live_url: Optional[str] = None,
 ) -> Optional[str]:
     """
     Set up LTTng session, with events and context.
@@ -193,31 +192,16 @@ def setup(
         the usual page size)
     :param subbuffer_size_kernel: the size of the subbuffers for kernel events (defaults to 32
         times the usual page size, since there can be way more kernel events than UST events)
-    :param live_mode: whether to create a live session
     :param live_timer_interval: the time interval at which the data should be flushed from the
         buffer and sent to the LTTng relay. This is in microseconds.
-        Used only if live_mode is `True`.
-    :param live_tracing_url_origin: the URL to which the tracing output will be sent.
-        Used only if live_mode is `True`.
+        Used only if live_timer_interval is `True`.
+    :param live_url: the URL to which the tracing output will be sent.
+        Used only if live_timer_interval is `True`.
     :return: the full path to the trace directory, or `None` if initialization failed
     """
     # Validate parameters
     if not session_name:
         raise RuntimeError('empty session name')
-    # Resolve full tracing directory path
-    # TODO(christophebedard): do we need to join the
-    #   base_path with session_name for a live session?
-    #   We need to return a path, so maybe format it like:
-    #   "net://localhost/host/$hostname/$session_name"
-    live_tracing_url = (
-        live_tracing_url_origin + '/' +
-        'host/' + socket.gethostname() + '/' +
-        session_name
-    )
-    full_path = os.path.join(base_path, session_name)
-    if os.path.isdir(full_path) and not append_trace:
-        raise RuntimeError(
-            f'trace directory already exists, use the append option to append to it: {full_path}')
 
     # If there is no session daemon running, try to spawn one
     if is_session_daemon_not_alive():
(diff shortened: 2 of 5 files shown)


christophebedard:
Thanks for the PR! I'll try to take a look tomorrow.

suchetanrs:
I just found out that a job that passed in my older PR now failed. Seems like a bunch of tests are failing :p 
I've managed to fix push a fix for few of them but I think most are still because of the API changes I made. 
I feel that I should attempt to fix this later because the API is still subject to change. Please let me know if this is okay! :)

christophebedard:
Yeah, don't worry, GitHub CI is a bit broken right now! See #160. You can ignore those test failures. I'll fix them once I get some time.

tracetools_trace/tracetools_trace/tools/args.py
```diff
@@ -80,12 +80,23 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
...
-        # Default value for 'lttng create-session --live':
-        # https://lttng.org/man/1/lttng-create/v2.13/#doc-opt--live
-        default=1000000,
-        help='TODO (default: %(default)s)')
+        '--live', dest='live_mode', action='store_true',
+        help='Enable live mode (default: %(default)s)')
+    parser.add_argument(
+        '--live-timer-interval', dest='live_timer_interval', type=int,
+        default=100000,
+        help='Set the live timer interval (default: %(default)s)')
```
christophebedard: You're totally right that the `--live` option didn't work correctly as I originally wrote it.

Fortunately, I figured out how to make it behave like I intended: we have to use [`const=`](https://docs.python.org/3/library/argparse.html#const) instead of `default=` to set the "default" value (`1000000`). Then we get the correct behaviour:

```console
$ ros2 trace  # don't enable live mode
userspace tracing enabled (40 events)
kernel tracing disabled
syscalls tracing disabled
context (3 fields)
writing tracing session to: /home/christophe.bedard/.ros/tracing/session-20250406165801
press enter to start...
live_timer_internal= None
$ ros2 trace --live  # enable live mode, default timer interval value
userspace tracing enabled (40 events)
kernel tracing disabled
syscalls tracing disabled
context (3 fields)
writing tracing session to: /home/christophe.bedard/.ros/tracing/session-20250406165806
press enter to start...
live_timer_internal= 1000000
$ ros2 trace --live 1234  # enable live mode, user-specified timer interval value
userspace tracing enabled (40 events)
kernel tracing disabled
syscalls tracing disabled
context (3 fields)
writing tracing session to: /home/christophe.bedard/.ros/tracing/session-20250406165811
press enter to start...
live_timer_internal= 1234
$ ros2 trace --live=5678  # this format also works
userspace tracing enabled (40 events)
kernel tracing disabled
syscalls tracing disabled
context (3 fields)
writing tracing session to: /home/christophe.bedard/.ros/tracing/session-20250406170258
press enter to start...
live_timer_internal= 5678
```

So we thankfully don't need a separate flag to enable/disable the live mode!

tracetools_trace/tracetools_trace/tools/args.py
```diff
@@ -80,12 +80,23 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
...
+    parser.add_argument(
+        '--live-tracing-url-origin', dest='live_tracing_url_origin', type=str,
+        default='net://localhost',
+        help='Set the live tracing URL origin (default: %(default)s)')
```
christophebedard: I agree, we need this option. I might just shorten the name to `'--live-url'`/`dest='live_url'`.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -190,15 +193,27 @@ def setup(
...
-    # TODO(christophebedard): do we need to join the base_path with session_name for a live session?
+    # TODO(christophebedard): do we need to join the
+    #   base_path with session_name for a live session?
     #   We need to return a path, so maybe format it like:
     #   "net://localhost/host/$hostname/$session_name"
+    live_tracing_url = (
+        live_tracing_url_origin + '/' +
+        'host/' + socket.gethostname() + '/' +
+        session_name
+    )
```
christophebedard: I actually don't yet fully understand what this URL represents. I think this works if you process the live trace from the same computer, but what should this URL be if you want to process the trace data from another computer and what URL should the other computer use? I think this is a pretty important use-case.

Would you be able to look into what the URL represents (if you haven't already)? If you have two computers on a local network that you can try this with, that would also help us understand. You could use the `bt2_lttng_live.py` script from #149 to receive trace data.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -342,6 +357,9 @@ def setup(
             context_fields=contexts_dict.get(domain),
         )
 
+    if live_mode:
+        # TODO(suchetanrs) is there a better way to do this?
+        full_path = live_tracing_url
```
christophebedard: You could possibly set `full_path` to `live_tracing_url` right before the call to `_create_session_live()`.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -190,15 +193,27 @@ def setup(
...
+    #   base_path with session_name for a live session?
     #   We need to return a path, so maybe format it like:
     #   "net://localhost/host/$hostname/$session_name"
+    live_tracing_url = (
+        live_tracing_url_origin + '/' +
+        'host/' + socket.gethostname() + '/' +
+        session_name
+    )
     full_path = os.path.join(base_path, session_name)
     if os.path.isdir(full_path) and not append_trace:
```
christophebedard: Here it's checking if the "trace directory" (`$base_path/$session_name`) already exists (because we raise an error if it exists and the user didn't explicitly want to re-use the same trace directory), but this probably doesn't make sense if we're using the live mode, since it's supposed to send the trace data somewhere else. You could move this check to right before the `_create_session()` call.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -440,7 +458,7 @@ def _create_session(
 def _create_session_live(
     *,
     session_name: str,
-    full_path: str,
+    live_tracing_url_origin: str,
```
christophebedard: The param name is a bit redundant inside this function, since this whole function is about live tracing. Maybe just `url`? To match `lttng_create_session_live()`'s `url` argument.

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -137,12 +139,19 @@ def init(
...
     if trace_directory is None:
         return False
     # Simple sanity check
-    assert trace_directory == full_session_path
+    print(f'Trace directory: {trace_directory}')
+    print(f'Full session path: {full_session_path}')
+    # TODO(suchetanrs): There should be a sanity check for live_mode as well.
+    # Should there be a _resolve_url_live just like  _resolve_session_path?
+    if not live_mode:
+        assert trace_directory == full_session_path
```
christophebedard: Good question. I don't think there is anything to resolve it to (i.e., no path), so this assertion indeed probably doesn't make sense when using the live mode.

Also, `_resolve_session_path()` (above) shouldn't be called when in live mode. We do need something that prints the URL when in live mode like `_resolve_session_path()` does for a normal trace directory, though. Something like:

```py
print(f'sending live trace data to: {live_tracing_url_origin}')
```

suchetanrs:
Thanks for your review comments!!
I will quickly setup a second computer with the tracing repository for the url comment and will work on all of them tomorrow once I finish my GSoC proposal for this :p

suchetanrs:
Hey! @christophebedard 

I've setup two different laptops with ros2 tracing. I can ssh between the laptops and ros communication between them seems to work. However, when I try to connect to the ip address using babeltrace, I always get a connection refused error. In fact, if I try to view the trace output from the same laptop for an ip that's not localhost, I get the same issue.

I've built LTTng from source and I'm trying to understand the flow of functions. It seems like the supported protocols from the input url are `net`, `net6`, `tcp` and `tcp6`. We can also specify the input ports we'd like to use (upto 2 ports for `net`/`net6` and a single one for `tcp`) in the following manner `protocol://address:PORT1:PORT2` (just sharing some useful info that I found here :p).
If I use the `tcp` protocol instead of `net`, I cannot view the output on babeltrace because it says `Unknown protocol: tcp6`. I'm not sure if babeltrace supports `tcp`.
I'm still looking for the exact location in the lttng code base where the input url is resolved to something like `net://localhost/host/host_name/session_name`. 

Please let me know if you have some inputs here. I'll continue to read and understand what's done in LTTng before finishing the PR just so I'm aware of what I'm doing :p

christophebedard:
Thanks for looking into it!

I think we might have to use the LTTng relay daemon (`lttng-relayd`) if we're sending the tracing data from one system to another. I found this example here: https://pavelmakhov.com/2017/01/lttng-streaming/ (note the comment about the ports needing to be open; not sure if that will be an issue for you, otherwise maybe you can try different ports?). So the URL on the system that is tracing should point to the host that is receiving the trace data/that has a relay daemon. The host that is receiving the trace data should start a relay daemon, and the URL used for receiving the trace data should point to localhost _but_ include the hostname of the remote system (see the last code block).

Knowing the above, these sections in the LTTng docs make a lot more sense:

1. The last paragraph of this section about live tracing and starting the relay daemon on another system: https://lttng.org/docs/v2.13/#doc-lttng-live
2. This whole section about sending trace data over the network and pointing to the host on which the relay daemon is running: https://lttng.org/docs/v2.13/#doc-sending-trace-data-over-the-network

If we combine those two use-cases, we basically achieve what we want, i.e., live tracing and sending the data to a remote system. And the URL format makes more sense!

Let me know if that makes sense and if you can get it to work knowing this.

> I've built LTTng from source

You're very courageous! :grin:

tracetools_trace/tracetools_trace/tools/args.py
suchetanrs: Thanks!! 
I've added this in the PR.

tracetools_trace/tracetools_trace/tools/args.py
suchetanrs: Done!

tracetools_trace/tracetools_trace/trace.py
suchetanrs: Understood!
I've changed this so that `_resolve_session_path` is called only when the mode is local mode. 
Since the url needs to have the host name of the system where the tracing is running, I think it's possible to compute it here itself. So I've called a function that does that here. The user can use this printed URL directly on the remote or local machine to view the trace.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done!

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done!

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done!

suchetanrs:
Thanks a lot for your reply @christophebedard !!
Sorry for the late reply. I completely missed this part in the documentation. When I was trying to read the trace data with `babeltrace`, instead of `localhost`, I always used the IP of the remote system. I guess that explains why I got the connection refused error. I followed the blog you shared and it seems to work well for me with two computers :smile: 

I went on an LTTng documentation study spree and hence the delay with PR. I've uploaded my study notes [here](https://drive.google.com/file/d/1cw6Ep-STvz5Uq6tbGVk1h0YNwkFkcjD4/view?usp=sharing) :p 

I've addressed your review comments but I have a question:
Do we expect users to start `lttng-relayd` manually? I'm not sure if it makes a lot of sense to do this but if the input url is just `localhost`, can we spawn the relayd on the same machine similar to the below code using `lttng-relayd --daemonize`? https://github.com/ros2/ros2_tracing/blob/aa718f3fdfdf8c4fd907cf33accf8e11914ba804/tracetools_trace/tracetools_trace/tools/lttng_impl.py#L134

If we do this, maybe we can somehow use the `--path` parameter to write the live traces to the default folder with `$ROS_HOME` instead of `$LTTNG_HOME`?

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: It worked well with 2 machines! :)
Explained further in https://github.com/ros2/ros2_tracing/pull/165#issuecomment-2798250016

christophebedard:
Thanks for the changes, I will take a look tomorrow.

> Sorry for the late reply. I completely missed this part in the documentation. When I was trying to read the trace data with babeltrace, instead of localhost, I always used the IP of the remote system. I guess that explains why I got the connection refused error. I followed the blog you shared and it seems to work well for me with two computers 😄

Great!

> I went on an LTTng documentation study spree and hence the delay with PR. I've uploaded my study notes [here](https://drive.google.com/file/d/1cw6Ep-STvz5Uq6tbGVk1h0YNwkFkcjD4/view?usp=sharing) :p

I'll have to take a look at this!

> Do we expect users to start lttng-relayd manually? I'm not sure if it makes a lot of sense to do this but if the input url is just localhost, can we spawn the relayd on the same machine similar to the below code using lttng-relayd --daemonize?

Oh, interesting! It does look like the `lttng create` command is the one that automatically starts the relay daemon if you create a `--live` session with the `lttng` CLI and don't provide a URL (and not liblttng-ctl, which `lttngpy` uses): https://github.com/lttng/lttng-tools/blob/e710f55dbf2b15e4874d38875338ac695a31cbfa/src/bin/lttng/commands/create.c#L255-L270

In that case, then yes we could definitely spawn it similar to `spawn_session_daemon()`, unless there's already a relay daemon.

suchetanrs:
Thanks!! :D 

I will add spawning the relayd to this as well!

tracetools_trace/tracetools_trace/tools/args.py
```diff
@@ -84,8 +84,12 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
         '--live', dest='live_timer_interval', type=int, nargs='?',
         # Default value for 'lttng create-session --live':
         # https://lttng.org/man/1/lttng-create/v2.13/#doc-opt--live
-        default=1000000,
-        help='TODO (default: %(default)s)')
+        const=100000,
+        help='Set the live timer interval (default: %(default)s)')
```
christophebedard: This does set the live timer interval, but the main use of the `--live` flag is to create a live tracing session, so this should be something like:

```suggestion
        help='Create a live tracing session. Optionally set the live timer interval (default: %(default)s)')
```

(double check the line length and split it over two lines if necessary)

tracetools_trace/tracetools_trace/tools/args.py
```diff
@@ -84,8 +84,12 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
...
+    parser.add_argument(
+        '--live-url', dest='live_url', type=str,
+        default='net://localhost',
+        help='Set the live tracing URL origin (default: %(default)s)')
```
christophebedard, line 93: This is just a note for some improvements later, no need to do anything now:

We should document what this all means (timer interval, URL, etc.), link to the relevant parts of the LTTng docs, provide some examples, etc. This could go in the README.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -190,19 +192,16 @@ def setup(
         the usual page size)
     :param subbuffer_size_kernel: the size of the subbuffers for kernel events (defaults to 32
         times the usual page size, since there can be way more kernel events than UST events)
+    :param live_timer_interval: the time interval at which the data should be flushed from the
+        buffer and sent to the LTTng relay. This is in microseconds.
```
christophebedard:
```suggestion
        buffer and sent to the LTTng relay daemon. This is in microseconds.
```

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -190,19 +192,16 @@ def setup(
         the usual page size)
     :param subbuffer_size_kernel: the size of the subbuffers for kernel events (defaults to 32
         times the usual page size, since there can be way more kernel events than UST events)
+    :param live_timer_interval: the time interval at which the data should be flushed from the
```
christophebedard, line 195: This should also say that the tracing session will be in live mode if this value is not `None`.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -190,19 +192,16 @@ def setup(
...
+        buffer and sent to the LTTng relay. This is in microseconds.
+        Used only if live_timer_interval is `True`.
```
christophebedard:
```suggestion
        Used only if live_timer_interval is not `None`.
```

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -190,19 +192,16 @@ def setup(
...
+    :param live_url: the URL to which the tracing output will be sent.
+        Used only if live_timer_interval is `True`.
```
christophebedard:
```suggestion
    :param live_url: the URL of the relay daemon to which the tracing output will be sent.
        Used only if live_timer_interval is `True`.
```

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -248,15 +247,26 @@ def setup(
...
         _create_session(
             session_name=session_name,
             full_path=full_path,
         )
     else:
+        # TODO(christophebedard): do we need to join the
+        #   base_path with session_name for a live session?
+        #   We need to return a path, so maybe format it like:
+        #   "net://localhost/host/$hostname/$session_name"
+        live_tracing_url =  'net://localhost/host/' + socket.gethostname() + '/' + session_name
```
christophebedard: Using f-strings here would be great instead of concatenating strings

christophebedard: This is tricky, because the first part of the URL (`net://localhost`) depends on the `live_url` parameter. This URL is printed to the terminal for users (similar to the trace directory path).

This will work if the URL is used from the remote system (receiving data), since the relay daemon is going to be local to that host. However, would this URL work (using `babeltrace2` or the Python script) from the target system (tracing & sending data) AND the remote system (receiving data, like you mentioned in https://github.com/ros2/ros2_tracing/pull/165#discussion_r2054230667) if we used `live_url` instead of `net://localhost`? You would have to set `live_url` to `net://localhost` if it's `None`, though.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -248,15 +247,26 @@ def setup(
...
         # LTTng will create the parent directories if needed
         _create_session(
             session_name=session_name,
             full_path=full_path,
         )
     else:
+        # TODO(christophebedard): do we need to join the
+        #   base_path with session_name for a live session?
+        #   We need to return a path, so maybe format it like:
+        #   "net://localhost/host/$hostname/$session_name"
```
christophebedard: You can remove this comment

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -451,7 +461,7 @@ def _create_session_live(
         # TODO(christophebedard): figure out what to provide here as the URL
         #   This depends on how we expect users to use live tracing
         #   See the documentation for the url param of lttng_create_session_live()
```
christophebedard: You can remove this comment

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -138,11 +158,13 @@ def init(
...
         context_fields=context_fields,
         live_timer_interval=live_timer_interval,
+        live_url=live_url,
     )
     if trace_directory is None:
         return False
     # Simple sanity check
-    assert trace_directory == full_session_path
+    if live_timer_interval is None:
+        assert trace_directory == full_session_path
```
christophebedard, line 167: The value returned by `lttng.lttng_init()` (`trace_directory`) should be equivalent to `full_live_url` if we're in live mode, so we should assert that they are equal too

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -79,6 +83,15 @@ def _resolve_session_path(
...
     return base_path, full_session_path
 
+def _resolve_session_live_url(
+    *,
+    session_name: str,
+) -> str:
+    live_tracing_url = (
+        'net://localhost/host/' + socket.gethostname() + '/' +
+        session_name
+    )
```
christophebedard: Similar to my other comment, this should probably use `live_url` instead of `net://localhost`.

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -79,6 +83,15 @@ def _resolve_session_path(
...
+    print(f'live trace data will be sent to: {live_tracing_url} on the system running lttng-relayd')
```
christophebedard: I would remove "on the system running lttng-relayd" but still mention the relay daemon before the `:`, like:

```suggestion
    print(f'live trace data will be sent to the relay daemon at: {live_tracing_url}')
```

christophebedard (commented):
Some more comments, but overall this is looking good

suchetanrs:
Hey, I tried a few things about your question in https://github.com/ros2/ros2_tracing/pull/165#discussion_r2061055123

If we consider the following scenario: The tracing is launched on a machine say `M_trace` with host name `M_trace_host` and IP `M_trace_ip` and the relay daemon is running on a machine on the same network say `M_remote` and `M_remote_host` and IP `M_remote_ip`.

At first, I started the relay daemon only using the command `lttng-relayd` on `M_remote`. In this case, the relay daemon listens for viewer connections only on `localhost`, so it would work only if I ran babeltrace on `M_remote` using the URL `net://localhost/host/M_trace_host/session-name`
In this setting, if I ran babeltrace on `M_trace` with the URL `net://localhost/host/M_trace_host/session-name`, I got a connection refused error without any surprises. Even with the URL `net://M_remote_ip/host/M_trace_host/session-name`, I got the same error.

On the other hand, if I start the relay daemon on `M_remote` using the command `lttng-relayd --live-port=tcp://0.0.0.0:5344` instead of just `lttng-relayd`, it enables the daemon to find viewer connections anywhere in the network https://lttng.org/man/8/lttng-relayd/v2.13/. So this time when I ran babeltrace with the URL `net://localhost/host/M_trace_host/session-name` or `net://M_remote_ip/host/M_trace_host/session-name` on `M_remote`, both worked fine. And on `M_trace`, running babeltrace with the URL `net://M_remote_ip/host/M_trace_host/session-name` worked fine. 

So the bottom line is that I am confused about what should we inform the user when they run live tracing :joy: 
If we say that the live trace output will be available at `net://localhost/host/M_trace_host/session-name` then it would enforce them to run the relay-d using `lttng-relayd` on the same machine as babeltrace otherwise it would not work. 

On the other hand, like you mentioned in https://github.com/ros2/ros2_tracing/pull/165#discussion_r2061059207, if we use the `live_url` directly, we should inform the user about running the relay daemon with the command `lttng-relayd --live-port=tcp://0.0.0.0:5344` so they can run babeltrace on any machine with our informed `live_url`. Shall I continue the PR with the later setting and adding a comment about starting the relay daemon with the extra --live-port?

tracetools_trace/tracetools_trace/tools/args.py
suchetanrs: Done.

tracetools_trace/tracetools_trace/tools/args.py
suchetanrs: Okay, understood :+1: 
I will raise a PR for the updates in the README after this.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done

suchetanrs: I have moved this to `live_url` like you mentioned. More info about this is here https://github.com/ros2/ros2_tracing/pull/165#issuecomment-2837939021. I think we should also inform the user about how start the relay daemon in this case.....

> You would have to set live_url to net://localhost if it's None, though.

I could not understand this...When will this case happen? If the user gives the `--live-url` argument without any value then it will not work...and there is a default value in `args.py`. Did I miss some other location?

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done

tracetools_trace/tracetools_trace/trace.py
suchetanrs: Done

tracetools_trace/tracetools_trace/trace.py
suchetanrs: Done

tracetools_trace/tracetools_trace/trace.py
suchetanrs: Done

tracetools_trace/tracetools_trace/tools/lttng_impl.py
suchetanrs: Done

Issue #166, opened 2025-04-15
test_ros2trace tests hang on nightly Linux jobs


christophebedard:
### Operating System:

Latest Linux (Ubuntu + RHEL) target platforms

### ROS version or commit hash:

Latest Rolling from source + Kilted (https://github.com/osrf/buildfarm-tools/issues/183#issuecomment-3532940789)

### RMW implementation (if applicable):

(not directly involved in this)

### Client library (if applicable):

(not directly involved in this)

### 'ros2 doctor --report' output

N/A

### Steps to reproduce issue

This can't really be reproduced locally, but it's been happening regularly in nightly Linux CI jobs:

1. https://ci.ros2.org/job/nightly_linux-rhel_release/2237/
2. https://ci.ros2.org/job/nightly_linux_debug/3391/
3. https://ci.ros2.org/job/nightly_linux-aarch64_release/3069/
4. https://ci.ros2.org/job/nightly_linux_release/3430/
5. https://ci.ros2.org/job/nightly_linux_repeated/3799/

Note that tests are run in parallel here.

### Expected behavior

Tests pass or time out after a reasonable amount of time.

### Actual behavior

`test_ros2trace` tests hang and never time out. There is no test result ("The test invocation failed without generating a result file.").

### Additional information

CLI test flakiness or hanging is definitely not something we've never seen. However, these specific tests could probably be improved.

It's hard to say where the tests are hanging. However, they do wait for a certain output here: https://github.com/ros2/ros2_tracing/blob/619b236277bda7aa4881cb27c88844cd37b61fe8/test_ros2trace/test/test_ros2trace/test_trace.py#L198-L203

We should look into other, more reliable ways to wait until `ros2 trace` has started tracing.

Overall, though, we could start by adding a timeout.

Crola1702:
This issue happened twice in a row in nightly_linux_repeated: 
- https://ci.ros2.org/job/nightly_linux_repeated/4012/
- https://ci.ros2.org/job/nightly_linux_repeated/4013/

And twice in a day with: https://ci.ros2.org/job/nightly_linux-rhel_repeated/2448/ (today)

christophebedard:
The hard part is that we're getting no logs/test output at all, at least not from the job/console output. Are we able to (manually) retrieve the test logs from the job, e.g., `log/latest_test/test_ros2trace/**`? Maybe the next time we see it hanging?

Crola1702:
[...]

PR #167, opened 2025-04-15
Add timeout to test_ros2trace tests that wait on stdout

christophebedard:
Part of #166

As mentioned in #166, I think the part that is most likely to hang is the part in `run_trace_command_start()` that waits for a specific output before proceeding. This is required in some cases because we want to make sure that tracing has started before launching nodes, for example.

Therefore, add an optional timeout, and use it wherever a caller/test wants to wait for tracing to start (`wait_for_start=True`).

I think this is a quick change that will at least help us rule this out as an issue. Long term, I'd like to use another mechanism to wait for tracing to start. I think we can have something like [`lttngpy::get_session_names()`](https://github.com/ros2/ros2_tracing/blob/14f7434e368f6d988524c5548cc0bae2e3229cbd/lttngpy/src/lttngpy/session.cpp#L27) that returns a list of tracing sessions _currently tracing_. We would just need to either explicitly set a tracing session name (which we don't always do, since there's a default session name) or be able to figure out the session name after the fact (may not be very reliable).

### DIFF
--- a/test_ros2trace/test/test_ros2trace/test_trace.py
+++ b/test_ros2trace/test/test_ros2trace/test_trace.py
@@ -16,6 +16,7 @@ import os
 import shutil
 import subprocess
 import tempfile
+import time
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -175,13 +176,15 @@ class TestROS2TraceCLI(unittest.TestCase):
     def wait_and_print_command_output(
         self,
         process: subprocess.Popen,
+        *,
+        timeout: Optional[float] = None,
     ) -> int:
         stdout, stderr = process.communicate()
         stdout = stdout.strip(' \r\n\t')
         stderr = stderr.strip(' \r\n\t')
         print('=>stdout:\n' + stdout)
         print('=>stderr:\n' + stderr)
-        return process.wait()
+        return process.wait(timeout)
 
     def run_trace_command_start(
         self,
@@ -189,6 +192,7 @@ class TestROS2TraceCLI(unittest.TestCase):
         *,
         env: Optional[Dict[str, str]] = None,
         wait_for_start: bool = False,
+        timeout: Optional[float] = None,
     ) -> subprocess.Popen:
         process = self.run_command(['ros2', 'trace', *args], env=env)
         # Write <enter> to stdin to start tracing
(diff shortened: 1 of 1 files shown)


christophebedard:
Pulls: ros2/ros2_tracing#167
Branch: christophebedard/test-ros2trace-wait-timeout
BUILD args:  --packages-above-and-dependencies test_ros2trace
TEST args:  --packages-above test_ros2trace
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15708
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23175/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17386/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2641/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24049/)

christophebedard:
@fujitatomoya would you be willing to review this? I'd like to merge it today/this week to prevent tests from hanging in nightly CI (#166).

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -199,37 +203,52 @@ def run_trace_command_start(
...
-            while 'press enter to stop...' not in stdout:
+            time_start = time.time()
+            time_now = time_start
+
+            def timedout(now: float) -> bool:
+                return timeout is not None and now - time_start > timeout
+
+            while 'press enter to stop...' not in stdout and not timedout(time_now):
                 stdout += process.stdout.read(1)
+                time_now = time.time()
```
fujitatomoya, line 214: maybe a bit simpler?

```suggestion
            time_start = time.time()

            def timedout() -> bool:
                return timeout is not None and time.time() - time_start > timeout

            while 'press enter to stop...' not in stdout and not timedout():
                stdout += process.stdout.read(1)
```

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -199,37 +203,52 @@ def run_trace_command_start(
...
+            self.assertFalse(
+                timedout(time_now),
```
fujitatomoya, line 216:
```suggestion
                timedout(),
```

fujitatomoya (approved):
lgtm with current PR, just a minor suggestion to make it simple.

test_ros2trace/test/test_ros2trace/test_trace.py
christophebedard: This could result in a false timeout if we exit from the `while` before it times out, but call `self.assertFalse(timedout(), ...)` after it times out.

The way I wrote it, that can't happen, since we set `time_now` before checking `timedout(time_now)` for the `while` and then don't re-set it before calling `self.assertFalse(timedout(time_now), ...)`.

christophebedard: @fujitatomoya does that make sense to you?

fujitatomoya: ah yeah, last `time_now` needs to be cached from above. that makes sense 👍

christophebedard:
Thanks for the review! Hopefully this stops tests from hanging.

PR #173, opened 2025-04-29
Warn if kernel might be paranoid about 'perf:thread:*' context fields

christophebedard:
This aims to warn users about a very specific situation that could lead to a trace being unexpectedly empty.

If the kernel is too paranoid about perf events (i.e., value in `/proc/sys/kernel/perf_event_paranoid` is too high) and [`CAP_SYS_ADMIN`](https://man7.org/linux/man-pages/man7/capabilities.7.html) is not available, then some perf events may not be available. See `perf_event_paranoid` in https://www.kernel.org/doc/Documentation/sysctl/kernel.txt.

In practice, this means that enabling the `perf:thread:task-clock` LTTng context field through `ros2 trace` (or the `Trace` action) would lead to no actual error and an unexpectedly-empty trace:

```
$ ros2 trace --list --context perf:thread:task-clock
```

In the end, this is an LTTng/perf/Linux kernel thing, but we can still try to detect the situation and print a warning so that users at least get a hint when they get an empty trace.

### DIFF
--- a/tracetools_trace/tracetools_trace/tools/lttng_impl.py
+++ b/tracetools_trace/tracetools_trace/tools/lttng_impl.py
@@ -15,13 +15,16 @@
 
 """Implementation of the interface for tracing with LTTng."""
 
+import itertools
 import os
+from pathlib import Path
 import shlex
 import subprocess
 from typing import Dict
 from typing import List
 from typing import Optional
 from typing import Set
+from typing import Tuple
 from typing import Union
 
 from lttngpy import impl as lttngpy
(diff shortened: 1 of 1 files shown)


christophebedard:
Funny thing is that there's a kernel patch which apparently Ubuntu and Debian have that introduces a `perf_event_paranoid` value of 4: https://github.com/torvalds/linux/commit/2986a639d3181e4686c5b88303a54f439d33c34a

The kernel on my system (Ubuntu 22.04, 5.15.0-134-generic) seems to have it:

```
$ cat /boot/config-$(uname -r) | grep SECURITY_PERF_EVENTS_RESTRICT
CONFIG_SECURITY_PERF_EVENTS_RESTRICT=y
```

fujitatomoya:
> Funny thing is that there's a kernel patch which apparently Ubuntu and Debian have that introduces a perf_event_paranoid value of 4: torvalds/linux@2986a63

yeah, depends on the platform. often done for enhanced security in environments like Android or Debian.

- Ubuntu 20/22 amd64

```console
tomoyafujita@~/DVT >cat /proc/sys/kernel/perf_event_paranoid
4
```

- Raspi4

```console
ubuntu@raspi4-1:~$ cat /proc/sys/kernel/perf_event_paranoid
3
```

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -55,6 +58,48 @@ def is_kernel_tracer_available() -> bool:
...
+    # For now, just check for 'perf:thread:*' perf events
+    has_perf_thread_context_fields = any(
+        # Wildcards (*) are not supported for context fields
+        context_field.startswith('perf:thread:')
+        for context_field in context_fields
+    )
+    if not has_perf_thread_context_fields:
+        return False, None
+    path_perf_event_paranoid = '/proc/sys/kernel/perf_event_paranoid'
+    paranoid_value_str = Path(path_perf_event_paranoid).read_text().strip()
```
fujitatomoya: i would add the try&catch statement here just in case? even with linux system, this depends on the kernel configuration i guess... i would catch `FileNotFoundError`, `PermissionError`.

fujitatomoya (approved):
just a minor comment that does not block this PR.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
christophebedard: Good point. Do you have any specific example of a Linux distro that doesn't contain a `/proc/sys/kernel/perf_event_paranoid` file? From what I can tell, it should always exist, although I guess kernels can always be patched and modified, so this could not be true. In that case, I can just assume that everything is good if the file doesn't exist (or if it's not a file).

I also removed a `print()` that I mistakenly left behind :laughing:

christophebedard:
Pulls: ros2/ros2_tracing#173
Gist: https://gist.githubusercontent.com/christophebedard/58db59fd0d6c157ba0e99a34f627c44c/raw/07ef21b691e6a8bd80bd7f4e40d931ccefcb285f/ros2.repos
BUILD args:  --packages-above-and-dependencies tracetools_trace
TEST args:  --packages-above tracetools_trace
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16063
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23728/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17832/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3172/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24431/)

tracetools_trace/tracetools_trace/tools/lttng_impl.py
fujitatomoya: what i only think of is that user intentionally disable instrument flag for kernel config. which is the case that user know instruments cannot be used, so maybe no need to catch the exception, i guess. maybe you can ignore this comment.

christophebedard: That's a good point, but then it becomes harder to detect. I think we can proceed with this, which should detect most cases, and improve it later if we think it's needed.

christophebedard:
Thanks for the review!

PR #175, opened 2025-05-03
Switch from master to main for setup-ros & action-ros-ci

christophebedard:
Relates to https://github.com/ros-tooling/setup-ros/pull/807  
Relates to https://github.com/ros-tooling/action-ros-ci/pull/974

### DIFF
--- a/.github/workflows/sanity.yml
+++ b/.github/workflows/sanity.yml
@@ -24,7 +24,7 @@ jobs:
     - uses: actions/checkout@v4
       with:
         path: ws/src/ros2/ros2_tracing
-    - uses: ros-tooling/setup-ros@master
+    - uses: ros-tooling/setup-ros@main
       with:
         required-ros-distributions: ${{ matrix.distro }}
         use-ros2-testing: true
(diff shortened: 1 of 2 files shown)


fujitatomoya (approved):
lgtm. if master does not exist any more, that will fail the CI.

btw, why do we use `main` instead of specific version? i think `main` is good for less maintenance, but fixed version could be more stable?

christophebedard:
> lgtm. if master does not exist any more, that will fail the CI.

We set up `main` -> `master` mirroring, but that's more of a transition helper; we should move to `main`.

> btw, why do we use main instead of specific version? i think main is good for less maintenance, but fixed version could be more stable?

To be honest, I've been intentionally using `@master` in my projects (including `ros2_tracing`) to always test the latest versions of `setup-ros`/`action-ros-ci` and catch issues as soon as possible. I've never really had issues, and actions in this repo are more of an additional check; what really matters is ci.ros2.org.

christophebedard:
I skipped ci.ros2.org because this only changes the GitHub CI config.

christophebedard:
@mergifyio backport kilted

PR #185, opened 2025-06-16
Add runtime tracing opt-out mechanism

### LINKED ISSUE #2860: `rclcpp` brings in `lttng-usr` threads by default
This isn't quite a bug report nor a feature request. We've noticed that in later ROS 2 distributions, `rclcpp` links in [`tracetools`](https://github.com/ros2/ros2_tracing) by default, which in turn links `lttng-usr`. `lttng-usr` has this fun habit of [spawning threads](https://github.com/lttng/lttng-ust/blob/611ccd9f2f6581faa48b47881520069c30c31daa/src/lib/lttng-ust/lttng-ust-comm.c#L2262-L2270) on `dl` load. That can be problematic when Linux namespaces and [unsharing](https://man7.org/linux/man-pages/man2/unshare.2.html) are in the mix.

Naively, I would imagine that only executables that you want to trace should link `lttng-usr`. `rclcpp` right now makes it so that every ROS 2 aware C++ executable will bring the `lttng` machinery with it.

Could we do better than that? Do we absolutely need to link `lttng-usr` to compile `rclcpp`?

hidmic:
## Description

This patch makes ROS 2 specific tracepoint providers dynamically loadable, and then uses this functionality to opt them out on `tracetools` dload.

Closes https://github.com/ros2/rclcpp/issues/2860.

Not for the most part, unless the `TRACETOOLS_DISABLE` envvar is set.

No.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ For more information about LTTng, [refer to its documentation](https://lttng.org
 
 ### Disabling the tracer at runtime
 
-To avoid loading the tracer at runtime (and therefore disable all instrumentation), set the `TRACETOOLS_RUNTIME_DISABLE` environment variable:
+To avoid loading the tracer at runtime (and therefore disable all instrumentation), set the `TRACETOOLS_RUNTIME_DISABLE` environment variable to `1`:
 
 ```
 $ export TRACETOOLS_RUNTIME_DISABLE=1
--- a/README.md
+++ b/README.md
@@ -100,9 +100,9 @@ $ sudo apt-get install lttng-modules-dkms
 
 For more information about LTTng, [refer to its documentation](https://lttng.org/docs/v2.13/).
 
-### Disabling the instrumentation
+### Disabling the tracer at runtime
 
-To disable all instrumentation in runtime, set the `TRACETOOLS_RUNTIME_DISABLE` envvar:
+To avoid loading the tracer at runtime (and therefore disable all instrumentation), set the `TRACETOOLS_RUNTIME_DISABLE` environment variable:
 
 ```
 $ export TRACETOOLS_RUNTIME_DISABLE=1
--- a/README.md
+++ b/README.md
@@ -100,6 +100,16 @@ $ sudo apt-get install lttng-modules-dkms
 
 For more information about LTTng, [refer to its documentation](https://lttng.org/docs/v2.13/).
 
+### Disabling the instrumentation
+
+To disable all instrumentation in runtime, set the `TRACETOOLS_RUNTIME_DISABLE` envvar:
+
+```
+$ export TRACETOOLS_RUNTIME_DISABLE=1
+$ ros2 run tracetools status
+Tracing disabled
+```
+
 ### Removing the instrumentation
 
 To build and remove all instrumentation, use `TRACETOOLS_DISABLED`:
--- a/test_ros2trace/test/test_ros2trace/test_trace.py
+++ b/test_ros2trace/test/test_ros2trace/test_trace.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections.abc import Mapping
+
 import os
 import shutil
 import subprocess
@@ -109,6 +111,13 @@ class TestROS2TraceCLI(unittest.TestCase):
 
     def assertTraceNotExist(self, trace_dir: str) -> None:
         self.assertFalse(os.path.isdir(trace_dir), f'trace directory exists: {trace_dir}')
+
+    def assertTraceIsEmpty(self, trace_dir: str) -> None:
+        self.assertTraceExist(trace_dir)
+        from tracetools_read.trace import get_trace_events
+        events_all = get_trace_events(trace_dir)
+        events = get_corresponding_trace_test_events(events_all, self.trace_test_id)
+        self.assertEqual(len(events), 0, f'events found: {events_all}')
 
     def assertTraceContains(
         self,
@@ -246,23 +255,25 @@ class TestROS2TraceCLI(unittest.TestCase):
         process = self.run_command(['ros2', 'trace', *args], env=env)
         return self.wait_and_print_command_output(process)
 
-    def run_nodes(self) -> None:
+    def run_nodes(self, env: Optional[Mapping[str, str]] = None) -> None:
         # Set trace test ID env var for spawned processes
-        env = os.environ.copy()
+        if env is None:
+            env = os.environ
+        mutable_env = dict(env)
         assert self.trace_test_id
-        env[TRACE_TEST_ID_ENV_VAR] = self.trace_test_id
+        mutable_env[TRACE_TEST_ID_ENV_VAR] = self.trace_test_id
         nodes = [
             Node(
                 package='test_tracetools',
                 executable='test_ping',
                 output='screen',
-                env=env,
+                env=mutable_env,
             ),
             Node(
                 package='test_tracetools',
                 executable='test_pong',
                 output='screen',
-                env=env,
+                env=mutable_env,
             ),
         ]
         ld = LaunchDescription(nodes)
@@ -635,5 +646,30 @@ class TestROS2TraceCLI(unittest.TestCase):
         ret = self.run_trace_subcommand(['stop', session_name])
         self.assertEqual(1, ret)
         self.assertTracingSessionNotExist(session_name)
+
+        shutil.rmtree(tmpdir)
+
+    @unittest.skipIf(not are_tracepoints_included(), 'tracepoints are required')
+    def test_runtime_disable(self) -> None:
+        tmpdir = self.create_test_tmpdir('test_runtime_disable')
+        session_name = 'test_runtime_disable'
+
+        env = os.environ.copy()
+        env['TRACETOOLS_RUNTIME_DISABLE'] = '1'
+
+        process = self.run_trace_command_start(
+            [
+                '--path', tmpdir, '--session-name', session_name,
+                '--ust', tracepoints.rcl_subscription_init, TRACE_TEST_ID_TP_NAME,
+            ],
+            wait_for_start=True,
+            env=env,
+        )
+        self.run_nodes(env)
+
+        ret = self.run_trace_command_stop(process)
+        self.assertEqual(0, ret)
+
+        self.assertTraceIsEmpty(os.path.join(tmpdir, session_name))
 
         shutil.rmtree(tmpdir)
--- a/test_ros2trace/test/test_ros2trace/test_trace.py
+++ b/test_ros2trace/test/test_ros2trace/test_trace.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 from collections.abc import Mapping
-
 import os
 import shutil
 import subprocess
@@ -36,6 +35,7 @@ from tracetools_test.mark_process import TRACE_TEST_ID_ENV_VAR
 from tracetools_test.mark_process import TRACE_TEST_ID_TP_NAME
 from tracetools_trace.tools import tracepoints
 from tracetools_trace.tools.lttng import is_lttng_installed
+from tracetools_trace.tools.names import DEFAULT_EVENTS_ROS
 
 
 def are_tracepoints_included() -> bool:
@@ -112,12 +112,21 @@ class TestROS2TraceCLI(unittest.TestCase):
     def assertTraceNotExist(self, trace_dir: str) -> None:
         self.assertFalse(os.path.isdir(trace_dir), f'trace directory exists: {trace_dir}')
 
-    def assertTraceIsEmpty(self, trace_dir: str) -> None:
+    def assertTraceNotContains(self, trace_dir: str, unexpected_event_names: List[str]) -> None:
         self.assertTraceExist(trace_dir)
         from tracetools_read.trace import get_trace_events
-        events_all = get_trace_events(trace_dir)
-        events = get_corresponding_trace_test_events(events_all, self.trace_test_id)
-        self.assertEqual(len(events), 0, f'events found: {events_all}')
+        events = get_trace_events(trace_dir)
+        trace_test_events = get_corresponding_trace_test_events(events, self.trace_test_id)
+        self.assertGreater(
+            len(trace_test_events),
+            0,
+            f'no matching trace test events found: {events}')
+        for event in events:
+            event_name = get_event_name(event)
+            self.assertNotIn(
+                event_name, unexpected_event_names,
+                f'{event_name} found in events: {events}'
+            )
 
     def assertTraceContains(
         self,
@@ -257,23 +266,23 @@ class TestROS2TraceCLI(unittest.TestCase):
 
     def run_nodes(self, env: Optional[Mapping[str, str]] = None) -> None:
         # Set trace test ID env var for spawned processes
-        if env is None:
-            env = os.environ
-        mutable_env = dict(env)
+        additional_env: Dict[str, str] = {}
+        if env is not None:
+            additional_env.update(env)
         assert self.trace_test_id
-        mutable_env[TRACE_TEST_ID_ENV_VAR] = self.trace_test_id
+        additional_env[TRACE_TEST_ID_ENV_VAR] = self.trace_test_id
         nodes = [
             Node(
                 package='test_tracetools',
                 executable='test_ping',
                 output='screen',
-                env=mutable_env,
+                additional_env=additional_env,
             ),
             Node(
                 package='test_tracetools',
                 executable='test_pong',
                 output='screen',
-                env=mutable_env,
+                additional_env=additional_env,
             ),
         ]
         ld = LaunchDescription(nodes)
@@ -654,8 +663,7 @@ class TestROS2TraceCLI(unittest.TestCase):
         tmpdir = self.create_test_tmpdir('test_runtime_disable')
         session_name = 'test_runtime_disable'
 
-        env = os.environ.copy()
-        env['TRACETOOLS_RUNTIME_DISABLE'] = '1'
+        env = {'TRACETOOLS_RUNTIME_DISABLE': '1'}
 
         process = self.run_trace_command_start(
             [
@@ -670,6 +678,6 @@ class TestROS2TraceCLI(unittest.TestCase):
         ret = self.run_trace_command_stop(process)
         self.assertEqual(0, ret)
 
-        self.assertTraceIsEmpty(os.path.join(tmpdir, session_name))
+        self.assertTraceNotContains(os.path.join(tmpdir, session_name), DEFAULT_EVENTS_ROS)
 
         shutil.rmtree(tmpdir)
--- a/test_tracetools/src/mark_process.cpp
+++ b/test_tracetools/src/mark_process.cpp
@@ -34,6 +34,11 @@ namespace
 /**
  * \see tracetools_test
  */
+constexpr std::string_view tracetools_runtime_disable_env_var = "TRACETOOLS_RUNTIME_DISABLE";
+
+/**
+ * \see tracetools_test
+ */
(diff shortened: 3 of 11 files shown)


hidmic:
@christophebedard PTAL.

FYI @frneer @calderpg.

hidmic:
@christophebedard friendly ping

christophebedard:
Thanks for the PR!

I need to schedule some time to take a proper look and see if there are any unintended consequences/downsides.

christophebedard:
In the meantime, could you rebase your branch on `rolling` to get rid of some unrelated GitHub CI test failures?

hidmic:
Done

tracetools/src/tracetools.c
```diff
@@ -489,6 +491,17 @@ DEFINE_TRACEPOINT(
   TRACEPOINT_ARGS(
     buffer))
 
+void __attribute__((constructor)) tracetools_init()
+{
+  if(getenv("TRACETOOLS_DISABLE")) {
```
christophebedard: I would give it a different name just to avoid any confusion with the existing `TRACETOOLS_DISABLED` CMake option/define. Maybe `TRACETOOLS_RUNTIME_DISABLE`?

```suggestion
  if(getenv("TRACETOOLS_RUNTIME_DISABLE")) {
```

The environment variable should also be documented. I wouldn't put it [under the "Building" section next to the documentation about `TRACETOOLS_DISABLED`/`TRACETOOLS_TRACEPOINTS_EXCLUDE`](https://github.com/ros2/ros2_tracing#building) since this isn't a build thing. Maybe as a new section after the ["Tracing" section](https://github.com/ros2/ros2_tracing#tracing)? For example, an "Opt-out from tracing at runtime" section where this mechanism and the env var are explained.

tracetools/src/tracetools.c
```diff
@@ -489,6 +491,17 @@ DEFINE_TRACEPOINT(
   TRACEPOINT_ARGS(
     buffer))
 
+void __attribute__((constructor)) tracetools_init()
```
christophebedard, line 506: We probably don't _need_ a shared lib destructor, but it might be a good idea: looks like lttng-ust has things to clean up https://github.com/lttng/lttng-ust/blob/83d51a7498a113adc2c124e502edfa717564f630/src/lib/lttng-ust/lttng-ust-comm.c#L2397-L2400

tracetools/src/tracetools.c
```diff
@@ -489,6 +491,17 @@ DEFINE_TRACEPOINT(
...
+{
+  if(getenv("TRACETOOLS_DISABLE")) {
+    fprintf(stderr, "ROS 2 tracing disabled\n");
+    return;
+  }
+  if(!dlopen("libtracetools_provider.so", RTLD_NOW | RTLD_GLOBAL)) {
+    fprintf(stderr, "Failed to load tracepoint provider for ROS 2 tracing, disabling\n");
```
christophebedard: I'm assuming you probably won't keep the prints here, but I'd like to have something like it and include this mechanism in the [status tool (`ros2 run tracetools status`)](https://github.com/ros2/ros2_tracing#building):

```
$ TRACETOOLS_DISABLE=y ros2 run tracetools status 
ROS 2 tracing disabled
ROS 2 tracing disabled
Tracing enabled
```

I know it's not fully representative (e.g., if you set `TRACETOOLS_DISABLE=y` for your app but not `ros2 run tracetools status`), but I think it would still be useful if it's used in the same environment as an application. What do you think?

tracetools/src/tracetools.c
christophebedard: I'd also like to see a simple integration test for the mechanism. Maybe just a simple test in [`test_ros2trace` like this one](https://github.com/ros2/ros2_tracing/blob/d96270a26da9e58f426884753dba0c7a91f84d4b/test_ros2trace/test/test_ros2trace/test_trace.py#L297) that tries to trace some nodes that were launched with the env var and checks that the resulting trace is empty? Unless you have a better idea.

christophebedard (commented):
Thanks for the PR and for your patience :sweat_smile: I have some initial comments and requests.

I requested an update to the README, but mentioning this mechanism in the ROS 2 docs would also be great, possibly in the [existing guide](https://docs.ros.org/en/rolling/How-To-Guides/Building-ROS-2-with-Tracing.html). I can probably do it myself after this PR, though.

tracetools/src/tracetools.c
hidmic: Hmm I think it makes sense for the `ros2 run tracetools status` to report on it. I also think it makes sense to log something for the user. `stderr` is perhaps a bit raw. I could try and use `rcutils` logging capabilities, and also expose some flag for `ros2 run tracetools status` to check and provide a better reporting. Let me give this a shot.

tracetools/src/tracetools.c
hidmic: That sounds reasonable.

tracetools/src/tracetools.c
hidmic: Hmm, no, I don't like the `rcutils` dependency. I'll find another way.

tracetools/src/tracetools.c
hidmic: Done in 814bdd421e14cb4a6a469719edbe1e39f96fab99

tracetools/src/tracetools.c
hidmic: Added a destructor in 814bdd421e14cb4a6a469719edbe1e39f96fab99

tracetools/src/tracetools.c
hidmic: Sticking to `fprintf` with a verbosity flag. See 814bdd421e14cb4a6a469719edbe1e39f96fab99.

tracetools/src/tracetools.c
hidmic: Added in 814bdd421e14cb4a6a469719edbe1e39f96fab99

hidmic:
@christophebedard friendly ping

christophebedard:
Some CI jobs are failing. Can you take a look? They're failing in non-default build configurations (e.g., `TRACETOOLS_DISABLED=OFF` but `TRACETOOLS_TRACEPOINTS_EXCLUDED=ON`), which seems to be related to some of these changes.

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections.abc import Mapping
+
```
christophebedard: Nitpick: Is the blank line needed here?

```suggestion
```

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -246,23 +255,25 @@ def run_trace_subcommand(
...
         assert self.trace_test_id
-        env[TRACE_TEST_ID_ENV_VAR] = self.trace_test_id
+        mutable_env[TRACE_TEST_ID_ENV_VAR] = self.trace_test_id
         nodes = [
             Node(
                 package='test_tracetools',
                 executable='test_ping',
                 output='screen',
-                env=env,
+                env=mutable_env,
```
christophebedard: Minor comment: Using `Node`'s (`ExecuteProcess`') `additional_env` option would probably make more sense here. I believe `run_trace_command_start()`, which is also called by your test and is provided with the same `env`, uses the `env` as an update to the current environment as well.

`TRACE_TEST_ID_ENV_VAR` could be moved to `additional_env=` at the same time.

tracetools/src/tracetools.c
```diff
@@ -489,6 +484,64 @@ DEFINE_TRACEPOINT(
...
+
+bool ros_trace_runtime_status(void)
+{
+  return tracetools_provider_handle != NULL;
+}
+
+static bool get_boolean_env(const char *name)
+{
+  const char * value = getenv(name);
+  return value != NULL && strcmp(value, "1") == 0;
```
christophebedard, line 503: Concerning this line here and in `mark_process.cpp`:

I'm kind of used to env vars enabling a functionality if they're simply set and non-empty, instead of specifically requiring something like `"1"`. What do you think?

tracetools/src/tracetools.c
christophebedard: Thanks!

test_tracetools/src/mark_process.cpp
```diff
@@ -42,13 +47,19 @@ void mark_trace_test_process()
 {
 #ifndef TRACETOOLS_DISABLED
   // See tracetools_test.mark_process for more details
-  const std::string env_var{trace_test_id_env_var};
-  const auto test_id = rcpputils::get_env_var(env_var.c_str());
+  const std::string runtime_disable_env_var{tracetools_runtime_disable_env_var};
+  const auto runtime_disable = rcpputils::get_env_var(runtime_disable_env_var.c_str());
+  if (!runtime_disable.empty() && runtime_disable == "1") {
+    return;
+  }
```
christophebedard: I'm not sure we should apply the mechanism here. This process marking thing is meant to be a test functionality, so it shouldn't necessarily be disabled if `TRACETOOLS_RUNTIME_DISABLE` is enabled.

Also, in that case, given the process marking functionality, the `test_runtime_disable` test in `test_ros2trace` doesn't actually test the runtime disabling mechanism, right? In practice, the test would still pass even if the nodes incorrectly generated trace events, simply because there would be no process with the given trace event markers. Therefore, there would be no trace events associated with this test as per `get_corresponding_trace_test_events()`.

Let me know if that makes sense.

tracetools/src/tracetools.c
christophebedard: `TRACETOOLS_VERBOSE` looks good.

> Hmm, no, I don't like the rcutils dependency. I'll find another way.

Agreed.

README.md
```diff
@@ -100,6 +100,16 @@ $ sudo apt-get install lttng-modules-dkms
 
 For more information about LTTng, [refer to its documentation](https://lttng.org/docs/v2.13/).
 
+### Disabling the instrumentation
+
+To disable all instrumentation in runtime, set the `TRACETOOLS_RUNTIME_DISABLE` envvar:
```
christophebedard: Also, I feel like "disable all instrumentation at runtime" isn't really what this mechanism does, and it might be confusing because of the typical LTTng (runtime) trace events configuration mechanism.

What about something closer to what you're describing in the PR description (and what you're actually trying to achieve)? Something like:

```suggestion
To avoid loading the tracer at runtime (and therefore disable all instrumentation), set the `TRACETOOLS_RUNTIME_DISABLE` environment variable:
```

README.md
```diff
@@ -100,6 +100,16 @@ $ sudo apt-get install lttng-modules-dkms
 
 For more information about LTTng, [refer to its documentation](https://lttng.org/docs/v2.13/).
 
+### Disabling the instrumentation
```
christophebedard: I would add "... at runtime"

Also, given my other comment about how to describe the feature in this section, I would probably change this to something like:

```suggestion
### Disabling the tracer at runtime
```

christophebedard (commented):
Some mostly-minor comments (aside from the non-default build configuration issues), thanks for your patience :sweat_smile:

tracetools/src/tracetools.c
hidmic: I don't have a strong opinion. Here, I just replicated the semantics of the old [`ROS_LOCALHOST_ONLY` envvar](https://github.com/ros2/rcl/blob/ca440022a4fec2ce4d3624e5412bd22bc73b247f/rcl/src/rcl/localhost.c#L44-L47). We can remove the equality requirement.

test_tracetools/src/mark_process.cpp
hidmic: Yeap, you are absolutely right. I clearly extrapolated from `get_corresponding_trace_test_events()`. The problem's with `assertTraceIsEmpty`, it is testing the wrong thing. See c117ed6f7976477bb014cf8292ccebf1c5456c87.

test_ros2trace/test/test_ros2trace/test_trace.py
hidmic: Indeed. I wanted to keep `env` behavior for both helpers, missed the `os.environ.copy()` before `subprocess.Popen`. Fixed in c117ed6f7976477bb014cf8292ccebf1c5456c87.

test_ros2trace/test/test_ros2trace/test_trace.py
hidmic: Nope, just an stylistic habit of mine. Removed in c117ed6f7976477bb014cf8292ccebf1c5456c87

README.md
hidmic: Reads nicely. Changed in c117ed6f7976477bb014cf8292ccebf1c5456c87

README.md
hidmic: Changed in c117ed6f7976477bb014cf8292ccebf1c5456c87

hidmic:
> Some CI jobs are failing. Can you take a look? They're failing in non-default build configurations (e.g., TRACETOOLS_DISABLED=OFF but TRACETOOLS_TRACEPOINTS_EXCLUDED=ON), which seems to be related to some of these changes.

Should be fixed now. A `target_compile_definitions()` macro was misplaced.

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -110,6 +112,22 @@ def assertTraceExist(self, trace_dir: str) -> None:
...
+        self.assertGreater(
+            len(trace_test_events),
+            0,
+            f'no matching trace test events found: {events}')
+        for event in events:
+            event_name = get_event_name(event)
+            self.assertNotIn(
+                event_name, unexpected_event_names,
+                f'{event_name} found in events: {events}'
+            )
```
christophebedard, line 129: Ah, `get_corresponding_trace_test_events()` probably shouldn't include the special `TRACE_TEST_ID_TP_NAME` marker events in the events it returns, which would allow you to simply check if the list of events in the marked processes is empty (similar to what you did before https://github.com/ros2/ros2_tracing/pull/185#discussion_r2213381564), but this is fine.

tracetools/src/tracetools.c
christophebedard: Yeah okay, we seem to use `=1` and `=0` in many places (e.g., [for logging env vars](https://docs.ros.org/en/rolling/Concepts/Intermediate/About-Logging.html#environment-variables)), and we even have `rcutils_get_env_var_zero_or_one()`. Let's keep this, but then I would make it explicit in the README:

```diff
-To avoid loading the tracer at runtime (and therefore disable all instrumentation), set the `TRACETOOLS_RUNTIME_DISABLE` environment variable:
+To avoid loading the tracer at runtime (and therefore disable all instrumentation), set the `TRACETOOLS_RUNTIME_DISABLE` environment variable to `1`:
```

hidmic: Done in https://github.com/ros2/ros2_tracing/pull/185/commits/9bba68582b295e1e13a3fa3549bdba540abb9386

hidmic:
@christophebedard is this one ready to go?

christophebedard (approved):
LGTM with green CI. Thanks for the PR and for patiently iterating.

christophebedard:
Pulls: ros2/ros2_tracing#185
Gist: https://gist.githubusercontent.com/christophebedard/67426a2e420b4caeafdf6cb4bf979f1f/raw/b3f70dc0f9f9701be094223c4d6b26dd4261f1b1/ros2.repos
BUILD args:  --packages-above-and-dependencies tracetools test_tracetools test_ros2trace
TEST args:  --packages-above tracetools test_tracetools test_ros2trace
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16644
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24512/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18560/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3966/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25057/)

christophebedard:
GitHub seems to be having some issues: https://www.githubstatus.com/. CI jobs are failing to download content from GitHub. I'll restart the jobs once that is resolved.

tracetools/src/tp_define.c
```diff
@@ -0,0 +1,17 @@
+// Copyright 2019 Robert Bosch GmbH
```
fujitatomoya, line 1: just checking, is this copyright correct?

fujitatomoya (approved):
just a minor comment for copyright, implementation looks good to me.

@christophebedard probably it would be nice to add this to ros2 documentation after this is merged?

tracetools/src/tp_define.c
christophebedard: I think this is because this file is basically a copy of another file (even though it's a tiny file), so the original copyright was kept. I think it's fine.

hidmic: It is a copy indeed.

hidmic:
> GitHub seems to be having some issues: https://www.githubstatus.com/. CI jobs are failing to download content from GitHub. I'll restart the jobs once that is resolved.

It seems we are past the flukes, can we retrigger CI?

christophebedard:
I just did :grin: I think the test failures on Linux-aarch64 and Linux-rhel are unrelated, so I didn't retrigger those.

christophebedard:
I've retried the Linux job a couple of times and it keeps failing during `launch_testing_ros` tests, but I can't see an actual test failure in the job output: https://ci.ros2.org/job/ci_linux/24512/consoleText

```
Failed   <<< launch_testing_ros [37.5s, exited with code -6]
```

It's a bit suspicious. Could you take a look?

hidmic:
> It's a bit suspicious. Could you take a look?

Just did. It's unrelated to this patch. It's a `rolling` issue, an `std::terminate` coming from `rclpy` C++ bindings. It can be seen on other CI jobs (e.g. https://ci.ros2.org/job/ci_linux/24514), though Jenkins is under reporting the failure because technically no test failed, it is the Python interpreter that crashes on exit.

I'll post an issue on `rclpy` as soon as I get a stacktrace out of my local build, but I think we can proceed with this patch (and a backport to Jazzy and Kilted please!).

Edit: https://github.com/ros2/rclpy/issues/1484 posted.

christophebedard:
Thanks for looking into it and opening an issue! 

> but I think we can proceed with this patch (and a backport to Jazzy and Kilted please!).

I _think_ we could, but I'm not 110% sure. I've added it to the PMC agenda for next week (August 5th).

ros-discourse:

https://discourse.openrobotics.org/t/ros-pmc-minutes-for-august-5-2025/49439/1

christophebedard:
> > but I think we can proceed with this patch (and a backport to Jazzy and Kilted please!).
> 
> I _think_ we could, but I'm not 110% sure. I've added it to the PMC agenda for next week (August 5th).

After considering it, I consider this a feature, and I wouldn't backport it to older distros.

hidmic:
@christophebedard I read the PMC meeting notes. What's the rationale for not backporting a feature that isn't API nor ABI incompatible? Maintenance cost?

hidmic:
Friendly ping @christophebedard

christophebedard:
> I read the PMC meeting notes. What's the rationale for not backporting a feature that isn't API nor ABI incompatible? Maintenance cost?

Risk. If it were a simple new API, it would be pretty risk-free, but it's not just a new function, right? I don't want to break something somehow, especially with what happened with `rmw_zenoh` recently. I know it _shouldn't_ affect end users because it mainly moves symbols that should be internal to `libtracetools.so` out to a separate `dlopen`ed lib, but it still seems risky.

Were you specifically looking to get a Jazzy backport?

PR #187, opened 2025-06-23
Make trace action parameters substitutable

### LINKED ISSUE #186: Make more parameters of the Trace action substitutable
### Description

A number of the `Trace` launch action arguments accept substitutions. See the arguments that use `SomeSubstitutionsType`: https://github.com/ros2/ros2_tracing/blob/70d3f70992c57f1a17e7efd009ae1bafd83a0245/tracetools_launch/tracetools_launch/action.py#L111-L122

Not all of them support substitutions, so modify them to support substitutions, namely: `append_timestamp`, `append_trace`, `subbuffer_size_ust`, `subbuffer_size_kernel`.

### Motivation

Flexibility. By supporting substitutions, these parameters can be controlled by a launch argument and can therefore [be set through the launch interface](https://docs.ros.org/en/rolling/Tutorials/Intermediate/Launch/Using-Substitutions.html#modifying-launch-arguments), e.g.:

```console
$ ros2 launch my_trace_launch.py subbuffer_size_ust:=123456789
```

They can also be controlled through an environment variable and any other substitution.

### Design / Implementation Considerations

**Consideration**: The parameters that don't currently support substitutions are `bool`s or `int`s, which is not part of `SomeSubstitutionsType`. Therefore, these would probably need to accept both `bool`/`int` and `SomeSubstitutionsType`, and we'd need to convert `bool`/`int` values to `str` and then call `normalize_to_list_of_substitutions()`. Then, when performing the substitutions later, we can just use `perform_typed_substitution()` to perform the substitutions and convert to the desired type, since substitutions usually get resolved to `str`.

Rough implementation steps:

1. Update `Trace.parse()` to parse substitutions
2. Update `Trace.__init__()` to accept substitutions and convert to `str` (if `bool`/`int`) and call `normalize_to_list_of_substitutions()`
3. Update return type of the corresponding `@property`s and update any relevant tests (since these properties will now return substitutions that need to be performed, instead of `bool`/`int`)
4. Perform substitutions in `Trace.execute()` using `perform_typed_substitution()`
5. Update tests to cover substitutions for these `Trace` parameters
    1. See this test: https://github.com/ros2/ros2_tracing/blob/70d3f70992c57f1a17e7efd009ae1bafd83a0245/test_tracetools_launch/test/test_tracetools_launch/test_trace_action.py#L244-L293
    2. That test only tests substitutions with the Python API, not through the XML or YAML frontends. We could add tests that use substitutions in the XML and YAML frontends. See these tests: https://github.com/ros2/ros2_tracing/blob/70d3f70992c57f1a17e7efd009ae1bafd83a0245/test_tracetools_launch/test/test_tracetools_launch/test_trace_action.py#L133-L187

### Additional Information

_No response_

ShravanDeva5327:
## Description

This PR updates the Trace action to make the parameters `append_timestamp`, `append_trace`, `subbuffer_size_ust` and `subbuffer_size_kernel`  to be specified using substitutions in Python launch files.

The related tests have been updated accordingly and are passing.

Fixes #186 

Yes.

No.

### Additional Information

Support for XML and YAML frontends is not yet implemented.

### DIFF
--- a/tracetools_launch/tracetools_launch/action.py
+++ b/tracetools_launch/tracetools_launch/action.py
@@ -127,8 +127,7 @@ class Trace(Action):
         """
         Create a Trace.
 
-        Substitutions are supported for the session name,
-        base path, and the lists of events and context fields.
+        Substitutions are supported for all parameters.
 
         For the lists of events, wildcards can be used, e.g., 'ros2:*' for
         all events from the 'ros2' tracepoint provider or '*' for all events.
--- a/tracetools_launch/tracetools_launch/action.py
+++ b/tracetools_launch/tracetools_launch/action.py
@@ -18,6 +18,7 @@
 import fnmatch
 import re
 import shlex
+from typing import cast
 from typing import Iterable
 from typing import List
 from typing import Mapping
@@ -161,7 +162,10 @@ class Trace(Action):
         self._session_name: List[Substitution] = [
             IfElseSubstitution(
                 str(append_timestamp) if isinstance(append_timestamp, bool)
-                else normalize_typed_substitution(append_timestamp, bool),
+                else cast(
+                    List[Substitution],
+                    normalize_typed_substitution(append_timestamp, bool)
+                ),
                 Trace.AppendTimestamp(session_name),
                 session_name
             )
--- a/tracetools_launch/tracetools_launch/action.py
+++ b/tracetools_launch/tracetools_launch/action.py
@@ -41,6 +41,7 @@ from launch.utilities import normalize_to_list_of_substitutions
 from launch.utilities import perform_substitutions
 from launch.utilities.type_utils import normalize_typed_substitution
 from launch.utilities.type_utils import perform_typed_substitution
+from launch.utilities.type_utils import NormalizedValueType
 from tracetools_trace.tools import lttng
 from tracetools_trace.tools import names
 from tracetools_trace.tools import path
@@ -200,7 +201,7 @@ class Trace(Action):
         return self._base_path
 
     @property
-    def append_trace(self) -> Union[bool, List[Substitution]]:
+    def append_trace(self) -> NormalizedValueType:
         return self._append_trace
 
     @property
@@ -226,11 +227,11 @@ class Trace(Action):
         return self._context_fields
 
     @property
-    def subbuffer_size_ust(self) -> Union[int, List[SomeSubstitutionsType]]:
+    def subbuffer_size_ust(self) -> NormalizedValueType:
         return self._subbuffer_size_ust
 
     @property
-    def subbuffer_size_kernel(self) -> Union[int, List[SomeSubstitutionsType]]:
+    def subbuffer_size_kernel(self) -> NormalizedValueType:
         return self._subbuffer_size_kernel
 
     @classmethod
--- a/tracetools_launch/tracetools_launch/action.py
+++ b/tracetools_launch/tracetools_launch/action.py
@@ -40,8 +40,8 @@ from launch.substitutions import TextSubstitution
 from launch.utilities import normalize_to_list_of_substitutions
 from launch.utilities import perform_substitutions
 from launch.utilities.type_utils import normalize_typed_substitution
-from launch.utilities.type_utils import perform_typed_substitution
 from launch.utilities.type_utils import NormalizedValueType
+from launch.utilities.type_utils import perform_typed_substitution
 from tracetools_trace.tools import lttng
 from tracetools_trace.tools import names
 from tracetools_trace.tools import path
--- a/tracetools_launch/tracetools_launch/action.py
+++ b/tracetools_launch/tracetools_launch/action.py
@@ -39,6 +39,8 @@ from launch.substitutions import IfElseSubstitution
 from launch.substitutions import TextSubstitution
 from launch.utilities import normalize_to_list_of_substitutions
 from launch.utilities import perform_substitutions
+from launch.utilities.type_utils import normalize_typed_substitution
+from launch.utilities.type_utils import perform_typed_substitution
 from tracetools_trace.tools import lttng
 from tracetools_trace.tools import names
 from tracetools_trace.tools import path
@@ -111,15 +113,15 @@ class Trace(Action):
         session_name: SomeSubstitutionsType,
         append_timestamp: bool = False,
         base_path: Optional[SomeSubstitutionsType] = None,
-        append_trace: bool = False,
+        append_trace: Union[bool, SomeSubstitutionsType] = False,
         events_ust: Iterable[SomeSubstitutionsType] = names.DEFAULT_EVENTS_ROS,
         events_kernel: Iterable[SomeSubstitutionsType] = [],
         syscalls: Iterable[SomeSubstitutionsType] = [],
         context_fields:
             Union[Iterable[SomeSubstitutionsType], Mapping[str, Iterable[SomeSubstitutionsType]]]
             = names.DEFAULT_CONTEXT,
-        subbuffer_size_ust: int = 8 * 4096,
-        subbuffer_size_kernel: int = 32 * 4096,
+        subbuffer_size_ust: Union[int, SomeSubstitutionsType] = 8 * 4096,
+        subbuffer_size_kernel: Union[int, SomeSubstitutionsType] = 32 * 4096,
         **kwargs,
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


tracetools_launch/tracetools_launch/action.py
```diff
@@ -195,7 +200,7 @@ def base_path(self) -> List[Substitution]:
         return self._base_path
 
     @property
-    def append_trace(self) -> bool:
+    def append_trace(self) -> Union[bool, List[Substitution]]:
```
christophebedard: The return type here (i.e., type of `self._append_trace`) should ideally be `NormalizedValueType`, which is the return type of `normalize_typed_substitution()` and is the type that `perform_typed_substitution()` accepts.

While `Union[bool, List[Substitution]]` is not functionally incorrect (if you look at `NormalizedValueType`, it does include `bool` and `List[Substitution]`), but it is not valid as far as `mypy` (the Python static type checker) is concerned. See the `mypy` test failures in `test_tracetools_launch`, e.g.:

```
test/test_tracetools_launch/test_trace_action.py: note: In member "_check_trace_action" of class "TestTraceAction":
test/test_tracetools_launch/test_trace_action.py:115:49: error: Argument 2 to "perform_typed_substitution" has incompatible type "int | list[str | Path | Substitution | Iterable[str | Path | Substitution]]"; expected "list[Substitution] | int | float | bool | str | list[int | list[Substitution]] | list[float | list[Substitution]] | list[bool | list[Substitution]] | list[str | list[Substitution]]"  [arg-type]
test/test_tracetools_launch/test_trace_action.py:119:49: error: Argument 2 to "perform_typed_substitution" has incompatible type "int | list[str | Path | Substitution | Iterable[str | Path | Substitution]]"; expected "list[Substitution] | int | float | bool | str | list[int | list[Substitution]] | list[float | list[Substitution]] | list[bool | list[Substitution]] | list[str | list[Substitution]]"  [arg-type]
```

Same for the other `@property` accessors that you're modifying.

tracetools_launch/tracetools_launch/action.py
```diff
@@ -158,7 +159,11 @@ def __init__(
         self._logger = logging.get_logger(__name__)
         self._session_name: List[Substitution] = [
             IfElseSubstitution(
-                str(append_timestamp), Trace.AppendTimestamp(session_name), session_name)
+                str(append_timestamp) if isinstance(append_timestamp, bool)
+                else normalize_typed_substitution(append_timestamp, bool),
```
christophebedard: This is an interesting case. `IfElseSubstitution()` accepts `SomeSubstitutionsType`, which is a subset of the `NormalizedValueType` returned by `normalize_typed_substitution()`: `normalize_typed_substitution()` can return a `bool` or a `List[Substitution]`, while `IfElseSubstitution()` accepts `List[Substitution]` but not `bool`. So I think this is correct, well done!

tracetools_launch/tracetools_launch/action.py
```diff
@@ -405,20 +411,26 @@ def execute(self, context: LaunchContext) -> List[Action]:
             if isinstance(self._context_fields, Mapping)
             else [perform_substitutions(context, field) for field in self._context_fields]
         )
+        subbuffersize_ust = perform_typed_substitution(context, self._subbuffer_size_ust, int)
+        subbuffersize_kernel = perform_typed_substitution(
+            context,
+            self._subbuffer_size_kernel,
+            int
+        )
```
christophebedard, line 424: Thanks for preserving the order everywhere!

tracetools_launch/tracetools_launch/action.py
```diff
@@ -221,11 +226,11 @@ def context_fields(
         return self._context_fields
 
     @property
-    def subbuffer_size_ust(self) -> int:
+    def subbuffer_size_ust(self) -> Union[int, List[SomeSubstitutionsType]]:
```
christophebedard: I think this should be:

```suggestion
    def subbuffer_size_ust(self) -> Union[int, List[Substitution]]:
```

right? Since `normalize_typed_substitution()` returns `List[Substitution]` or `int`. `SomeSubstitutionsType` is pre-normalization.

tracetools_launch/tracetools_launch/action.py
```diff
@@ -221,11 +226,11 @@ def context_fields(
...
         return self._subbuffer_size_ust
 
     @property
-    def subbuffer_size_kernel(self) -> int:
+    def subbuffer_size_kernel(self) -> Union[int, List[SomeSubstitutionsType]]:
```
christophebedard: same as above:

```suggestion
    def subbuffer_size_kernel(self) -> Union[int, List[Substitution]]:
```

christophebedard (commented):
Thanks for the PR! It overall looks great, I just have some minor comments.

christophebedard:
I think the session name parameter in frontends might not correctly support substitutions, see `Trace.parse()`: https://github.com/ros2/ros2_tracing/blob/70d3f70992c57f1a17e7efd009ae1bafd83a0245/tracetools_launch/tracetools_launch/action.py#L297

so that might be a bit confusing when figuring out how to add support for substitutions to frontends for these parameters. It should probably look like this (which is it's done for `base_path`): https://github.com/ros2/launch_ros/blob/747b6038c42dec6afcb4a3fa64acd286a781aeb0/launch_ros/launch_ros/actions/node.py#L316-L318

`test_tracetools_launch/test/test_tracetools_launch/test_trace_action.py` doesn't test substitutions in XML/YAML; it would be worth testing that. You can look at these for examples of frontend tests with substitutions:

* https://github.com/ros2/launch/blob/6854fae626f2948139e1734e68d712cc1e0def1b/launch_xml/test/launch_xml/test_for_loop.py
* https://github.com/ros2/launch/blob/6854fae626f2948139e1734e68d712cc1e0def1b/launch_yaml/test/launch_yaml/test_for_loop.py

christophebedard:
You could just focus on addressing my comments for this PR, and work on supporting substitutions in frontends in another PR. Just to keep it simple!

tracetools_launch/tracetools_launch/action.py
ShravanDeva5327: done!

tracetools_launch/tracetools_launch/action.py
ShravanDeva5327: ah, yes! But should I change it to `NormalizedTypeValue` instead, like you suggested for `append_trace`?

tracetools_launch/tracetools_launch/action.py
```diff
@@ -161,7 +162,7 @@ def __init__(
         self._session_name: List[Substitution] = [
             IfElseSubstitution(
                 str(append_timestamp) if isinstance(append_timestamp, bool)
-                else normalize_typed_substitution(append_timestamp, bool),
+                else cast(List[Substitution], normalize_typed_substitution(append_timestamp, bool)),
```
ShravanDeva5327: Now that I've imported `NormalizedValueType`, Clang started showing warnings about the type, so I'm using `cast` to suppress them.

tracetools_launch/tracetools_launch/action.py
christophebedard: oops, yes!

tracetools_launch/tracetools_launch/action.py
christophebedard: yeah, I think `mypy` isn't smart enough/can't really tell that `append_timestamp` can only be a substitution in the `else` here

christophebedard:
Pulls: ros2/ros2_tracing#187
Gist: https://gist.githubusercontent.com/christophebedard/a66d663ccaec0a1f23fcfe0115c991c7/raw/ba4947dc5d60178e9d4ea3a77f6004a13d8c9884/ros2.repos
BUILD args:  --packages-above-and-dependencies test_tracetools_launch tracetools_launch
TEST args:  --packages-above test_tracetools_launch tracetools_launch
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16302
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24067/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18139/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/3471/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24693/)

christophebedard (approved):
Looks good! Now we just need to wait for CI.

ros-discourse:

ros-discourse:

PR #191, opened 2025-07-21
Add support for starting tracing at runtime

### LINKED ISSUE #44: Find solution to allow starting to trace from any point in time

### LINKED ISSUE #190: Enable tracing after application launch using Dual Sessions approach
Related to #44

### **Implementation**

- The `Trace` action should be modified to accept a new parameter that allows the user to specify whether tracing should start before application launch or after some time.

- If the user chooses to start tracing after some time, the `Trace` action should:
  - Start a **snapshot session** at launch with the name `session_name_init`.
  - Set the output location of the snapshot session to  `~/.ros/tracing/session_name/`.
  - The user will later execute the `ros2 trace` command (with the same `session_name`) to start the runtime session, which will have the same output location `~/.ros/tracing/session_name/`.
  - All `ros2 trace` verbs (such as `pause`, `resume`, etc.) should be updated accordingly, if needed.

- If the user does not specify delayed tracing, tracing should start before application launch as it currently does.

We may consider defaulting the `session_name` to the launch file name (needs confirmation).

We may also add a mechanism that allows the `ros2 trace` command to automatically detect active snapshot sessions and set the `session_name` for the runtime session accordingly.

**This mechanism will ensure that no trace is written to disk unless the user explicitly starts tracing, while still storing the initialization events in sub-buffers.**

ShravanDeva5327:
## Description

At present, users must start tracing before launching their application in order to collect meaningful trace data, due to the way the ROS 2 source code is instrumented (See #190). 

This PR adds the ability to start tracing at runtime, provided some pre-configuration is done. Specifically, the `dual_session` argument of the `Trace` action must be set to `True`:

Setting the `dual_session` argument of Trace Action to `True`
```python
Trace(
    session_name='my-tracing-session',
    dual_session=True,
),
```
This starts an [LTTng session in snapshot mode](https://lttng.org/docs/v2.13/#doc-tracing-session-mode), which captures initialization trace data and makes it available for writing to disk on demand.

Once pre-configured, tracing can be started at runtime using the `ros2 trace` command with the `--dual-session` option, which records a snapshot from the pre-configured session and starts a runtime tracing session to record the runtime trace data.

Example:
```bash
ros2 trace --session-name my-tracing-session --dual-session
```
⚠️ The session name used here must match the one specified in the Trace action.

When tracing is started, data will be written under:
```
~/.ros/tracing/my-session-name/
├── runtime # trace data of runtime events
└── snapshot # trace data of initialization events
```
If tracing is never started, no data will be written to disk.

Fixes #190 
Fixes #44

### DIFF
--- a/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
+++ b/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
@@ -67,6 +67,13 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
     py::arg("session_name"),
     py::arg("url"));
   m.def(
+    "lttng_create_session_snapshot",
+    &lttng_create_session_snapshot,
+    "Create snapshot session.",
+    py::kw_only(),
+    py::arg("session_name"),
+    py::arg("url"));
+  m.def(
     "lttng_destroy_session",
     &lttng_destroy_session,
     "Destroy session.",
--- a/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
+++ b/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
@@ -25,6 +25,7 @@
 #  include "lttngpy/event.hpp"
 #  include "lttngpy/lttng.hpp"
 #  include "lttngpy/session.hpp"
+#  include "lttngpy/snapshot.hpp"
 #endif  // LTTNGPY_DISABLED
 #include "lttngpy/status.hpp"
 
@@ -97,6 +98,17 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
     "Stop tracing.",
     py::kw_only(),
     py::arg("session_name"));
+  m.def(
+    "lttng_record_snapshot",
+    &lttngpy::lttng_record_snapshot,
+    "Record a snapshot session.",
+    py::kw_only(),
+    py::arg("session_name"),
+    py::arg("id"),
+    py::arg("max_size"),
+    py::arg("name"),
+    py::arg("url"),
+    py::arg("wait"));
 
   // Session info
   m.def(
--- a/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
+++ b/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
@@ -75,6 +75,15 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
     py::arg("session_name"),
     py::arg("url"));
   m.def(
+    "lttng_add_snapshot_output",
+    &lttngpy::lttng_add_snapshot_output,
+    "Add an output object to the snapshot session.",
+    py::kw_only(),
+    py::arg("session_name"),
+    py::arg("max_size"),
+    py::arg("name"),
+    py::arg("url"));
+  m.def(
     "lttng_destroy_session",
     &lttng_destroy_session,
     "Destroy session.",
@@ -103,11 +112,7 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
     &lttngpy::lttng_record_snapshot,
     "Record a snapshot session.",
     py::kw_only(),
-    py::arg("session_name"),
-    py::arg("id"),
-    py::arg("max_size"),
-    py::arg("name"),
-    py::arg("url"));
+    py::arg("session_name"));
 
   // Session info
   m.def(
--- a/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
+++ b/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
@@ -75,8 +75,8 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
     py::arg("session_name"),
     py::arg("url"));
   m.def(
-    "lttng_add_snapshot_output",
-    &lttngpy::lttng_add_snapshot_output,
+    "add_snapshot_output",
+    &lttngpy::add_snapshot_output,
     "Add an output object to the snapshot session.",
     py::kw_only(),
     py::arg("session_name"),
@@ -108,8 +108,8 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
     py::kw_only(),
     py::arg("session_name"));
   m.def(
-    "lttng_record_snapshot",
-    &lttngpy::lttng_record_snapshot,
+    "record_snapshot",
+    &lttngpy::record_snapshot,
     "Record a snapshot session.",
     py::kw_only(),
     py::arg("session_name"));
--- a/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
+++ b/lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
@@ -107,8 +107,7 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
     py::arg("id"),
     py::arg("max_size"),
     py::arg("name"),
-    py::arg("url"),
-    py::arg("wait"));
+    py::arg("url"));
 
   // Session info
   m.def(
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -1,0 +1,37 @@
+#include <lttng/snapshot.h>
+#include <lttng/lttng-error.h>
+
+#include <string>
+
+#include "lttngpy/snapshot.hpp"
+
+namespace lttngpy
+{
+
+int lttng_record_snapshot(
+  const std::string & session_name,
+  const uint32_t id,
+  const uint64_t max_size,
+  const std::string & name,
+  const std::string & url,
+  const int wait)
+{
+  struct lttng_snapshot_output * output = lttng_snapshot_output_create();
+  if (nullptr == output) {
+    return -LTTNG_ERR_UNK;
+  }
+
+  // Set snapshot output attributes.
+  lttng_snapshot_output_set_id(id, output);
+  lttng_snapshot_output_set_size(max_size, output);
+  lttng_snapshot_output_set_name(name.c_str(), output);  
+  lttng_snapshot_output_set_local_path(url.c_str(), output);
+
+  // Add snapshot output to the session.
+  lttng_snapshot_add_output(session_name.c_str(), output);
+
+  int ret = lttng_snapshot_record(session_name.c_str(), output, wait);
+  return ret;
+}
+
+}  // namespace lttngpy
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -22,9 +22,8 @@
 namespace lttngpy
 {
 
-int lttng_record_snapshot(
+int lttng_add_snapshot_output(
   const std::string & session_name,
-  const uint32_t id,
   const uint64_t max_size,
   const std::string & name,
   const std::string & url)
@@ -45,16 +44,18 @@ int lttng_record_snapshot(
     };
 
   // Set snapshot output attributes.
-  if (!try_or_cleanup(lttng_snapshot_output_set_id(id, output))) {return ret;}
   if (!try_or_cleanup(lttng_snapshot_output_set_size(max_size, output))) {return ret;}
   if (!try_or_cleanup(lttng_snapshot_output_set_name(name.c_str(), output))) {return ret;}
   if (!try_or_cleanup(lttng_snapshot_output_set_local_path(url.c_str(), output))) {return ret;}
 
-  // Add snapshot output to the session.
-  if (!try_or_cleanup(lttng_snapshot_add_output(session_name.c_str(), output))) {return ret;}
+  // Add output object to the session.
+  ret = lttng_snapshot_add_output(session_name.c_str(), output);
+  return ret;
+}
 
-  // Record the snapshot.
-  ret = lttng_snapshot_record(session_name.c_str(), output, 1);
+int lttng_record_snapshot(const std::string & session_name)
+{
+  int ret = lttng_snapshot_record(session_name.c_str(), NULL, 1);
   return ret;
 }
 
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -24,7 +24,7 @@ int lttng_record_snapshot(
   // Set snapshot output attributes.
   lttng_snapshot_output_set_id(id, output);
   lttng_snapshot_output_set_size(max_size, output);
-  lttng_snapshot_output_set_name(name.c_str(), output);  
+  lttng_snapshot_output_set_name(name.c_str(), output);
   lttng_snapshot_output_set_local_path(url.c_str(), output);
 
   // Add snapshot output to the session.
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -36,22 +36,22 @@ int lttng_record_snapshot(
 
   int ret = 0;
   auto try_or_cleanup = [&](int function_call) -> bool {
-    ret = function_call;
-    if (0 != ret) {
-      lttng_snapshot_output_destroy(output);
-      return false;
-    }
-    return true;
-  };
+      ret = function_call;
+      if (0 != ret) {
+        lttng_snapshot_output_destroy(output);
+        return false;
+      }
+      return true;
+    };
 
   // Set snapshot output attributes.
-  if (!try_or_cleanup(lttng_snapshot_output_set_id(id, output))) return ret;
-  if (!try_or_cleanup(lttng_snapshot_output_set_size(max_size, output))) return ret;
-  if (!try_or_cleanup(lttng_snapshot_output_set_name(name.c_str(), output))) return ret;
-  if (!try_or_cleanup(lttng_snapshot_output_set_local_path(url.c_str(), output))) return ret;
+  if (!try_or_cleanup(lttng_snapshot_output_set_id(id, output))) {return ret;}
+  if (!try_or_cleanup(lttng_snapshot_output_set_size(max_size, output))) {return ret;}
+  if (!try_or_cleanup(lttng_snapshot_output_set_name(name.c_str(), output))) {return ret;}
+  if (!try_or_cleanup(lttng_snapshot_output_set_local_path(url.c_str(), output))) {return ret;}
 
   // Add snapshot output to the session.
-  if (!try_or_cleanup(lttng_snapshot_add_output(session_name.c_str(), output))) return ret;
+  if (!try_or_cleanup(lttng_snapshot_add_output(session_name.c_str(), output))) {return ret;}
 
   // Record the snapshot.
   ret = lttng_snapshot_record(session_name.c_str(), output, 1);
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -22,7 +22,7 @@
 namespace lttngpy
 {
 
-int lttng_add_snapshot_output(
+int add_snapshot_output(
   const std::string & session_name,
   const uint64_t max_size,
   const std::string & name,
@@ -34,7 +34,7 @@ int lttng_add_snapshot_output(
   }
 
   int ret = 0;
-  auto try_or_cleanup = [&](int function_call) -> bool {
+  auto try_or_cleanup = [&ret, output](int function_call) -> bool {
       ret = function_call;
       if (0 != ret) {
         lttng_snapshot_output_destroy(output);
@@ -53,10 +53,9 @@ int lttng_add_snapshot_output(
   return ret;
 }
 
-int lttng_record_snapshot(const std::string & session_name)
+int record_snapshot(const std::string & session_name)
 {
-  int ret = lttng_snapshot_record(session_name.c_str(), NULL, 1);
-  return ret;
+  return lttng_snapshot_record(session_name.c_str(), NULL, 1);  // The last parameter is unused
 }
 
 }  // namespace lttngpy
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -34,16 +34,27 @@ int lttng_record_snapshot(
     return -LTTNG_ERR_UNK;
   }
 
+  int ret = 0;
+  auto try_or_cleanup = [&](int function_call) -> bool {
+    ret = function_call;
+    if (0 != ret) {
+      lttng_snapshot_output_destroy(output);
+      return false;
+    }
+    return true;
+  };
+
   // Set snapshot output attributes.
-  lttng_snapshot_output_set_id(id, output);
-  lttng_snapshot_output_set_size(max_size, output);
-  lttng_snapshot_output_set_name(name.c_str(), output);
-  lttng_snapshot_output_set_local_path(url.c_str(), output);
+  if (!try_or_cleanup(lttng_snapshot_output_set_id(id, output))) return ret;
+  if (!try_or_cleanup(lttng_snapshot_output_set_size(max_size, output))) return ret;
+  if (!try_or_cleanup(lttng_snapshot_output_set_name(name.c_str(), output))) return ret;
+  if (!try_or_cleanup(lttng_snapshot_output_set_local_path(url.c_str(), output))) return ret;
 
   // Add snapshot output to the session.
-  lttng_snapshot_add_output(session_name.c_str(), output);
+  if (!try_or_cleanup(lttng_snapshot_add_output(session_name.c_str(), output))) return ret;
 
-  int ret = lttng_snapshot_record(session_name.c_str(), output, 1);
+  // Record the snapshot.
+  ret = lttng_snapshot_record(session_name.c_str(), output, 1);
   return ret;
 }
 
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -27,8 +27,7 @@ int lttng_record_snapshot(
   const uint32_t id,
   const uint64_t max_size,
   const std::string & name,
-  const std::string & url,
-  const int wait)
+  const std::string & url)
 {
   struct lttng_snapshot_output * output = lttng_snapshot_output_create();
   if (nullptr == output) {
@@ -44,7 +43,7 @@ int lttng_record_snapshot(
   // Add snapshot output to the session.
   lttng_snapshot_add_output(session_name.c_str(), output);
 
-  int ret = lttng_snapshot_record(session_name.c_str(), output, wait);
+  int ret = lttng_snapshot_record(session_name.c_str(), output, 1);
   return ret;
 }
 
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -46,7 +46,9 @@ int add_snapshot_output(
   // Set snapshot output attributes.
   if (!try_or_cleanup(lttng_snapshot_output_set_size(max_size, output))) {return ret;}
   if (!try_or_cleanup(lttng_snapshot_output_set_name(name.c_str(), output))) {return ret;}
-  if (!try_or_cleanup(lttng_snapshot_output_set_local_path(url.c_str(), output))) {return ret;}
+  if (!try_or_cleanup(lttng_snapshot_output_set_ctrl_url(("file://" + url).c_str(), output))) {
+    return ret;
+  }
 
   // Add output object to the session.
   ret = lttng_snapshot_add_output(session_name.c_str(), output);
--- a/lttngpy/src/lttngpy/snapshot.cpp
+++ b/lttngpy/src/lttngpy/snapshot.cpp
@@ -1,3 +1,17 @@
+// Copyright 2025 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 #include <lttng/snapshot.h>
 #include <lttng/lttng-error.h>
 
--- a/lttngpy/src/lttngpy/snapshot.hpp
+++ b/lttngpy/src/lttngpy/snapshot.hpp
@@ -22,7 +22,7 @@ namespace lttngpy
 
 /**
  * Record a snapshot session.
- * 
+ *
  * \param session_name the session name
  * \param id the snapshot output ID
  * \param max_size the maximum size of the snapshot output in bytes
--- a/lttngpy/src/lttngpy/snapshot.hpp
+++ b/lttngpy/src/lttngpy/snapshot.hpp
@@ -1,0 +1,28 @@
+#ifndef LTTNGPY__SNAPSHOT_HPP_
+#define LTTNGPY__SNAPSHOT_HPP_
+
+#include <string>
+
+namespace lttngpy
+{
+
+/**
+ * Record a snapshot session.
+ * 
+ * \param session_name the session name
+ * \param id the snapshot output ID
+ * \param max_size the maximum size of the snapshot output in bytes
+ * \param name the name of the snapshot output
+ * \param url the destination URL for the snapshot output
+ */
+int lttng_record_snapshot(
+  const std::string & session_name,
+  const uint32_t id,
+  const uint64_t max_size,
+  const std::string & name,
+  const std::string & url,
+  const int wait);
+
+} // namespace lttngpy
+
+#endif  // LTTNGPY__SNAPSHOT_HPP_
--- a/lttngpy/src/lttngpy/snapshot.hpp
+++ b/lttngpy/src/lttngpy/snapshot.hpp
@@ -23,6 +23,6 @@ int lttng_record_snapshot(
   const std::string & url,
   const int wait);
 
-} // namespace lttngpy
+}  // namespace lttngpy
 
 #endif  // LTTNGPY__SNAPSHOT_HPP_
--- a/lttngpy/src/lttngpy/snapshot.hpp
+++ b/lttngpy/src/lttngpy/snapshot.hpp
@@ -1,3 +1,17 @@
+// Copyright 2025 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 #ifndef LTTNGPY__SNAPSHOT_HPP_
 #define LTTNGPY__SNAPSHOT_HPP_
 
--- a/lttngpy/src/lttngpy/snapshot.hpp
+++ b/lttngpy/src/lttngpy/snapshot.hpp
@@ -34,8 +34,7 @@ int lttng_record_snapshot(
   const uint32_t id,
   const uint64_t max_size,
   const std::string & name,
-  const std::string & url,
-  const int wait);
+  const std::string & url);
 
 }  // namespace lttngpy
 
--- a/lttngpy/src/lttngpy/snapshot.hpp
+++ b/lttngpy/src/lttngpy/snapshot.hpp
@@ -21,20 +21,25 @@ namespace lttngpy
 {
 
 /**
- * Record a snapshot session.
+ * Add an output object to the snapshot session.
  *
  * \param session_name the session name
- * \param id the snapshot output ID
  * \param max_size the maximum size of the snapshot output in bytes
  * \param name the name of the snapshot output
  * \param url the destination URL for the snapshot output
  */
-int lttng_record_snapshot(
+int lttng_add_snapshot_output(
   const std::string & session_name,
-  const uint32_t id,
   const uint64_t max_size,
   const std::string & name,
   const std::string & url);
+
+/**
+ * Record a snapshot session.
+ *
+ * \param session_name the session name
+ */
+int lttng_record_snapshot(const std::string & session_name);
 
 }  // namespace lttngpy
 
--- a/lttngpy/src/lttngpy/snapshot.hpp
+++ b/lttngpy/src/lttngpy/snapshot.hpp
@@ -27,8 +27,9 @@ namespace lttngpy
  * \param max_size the maximum size of the snapshot output in bytes
  * \param name the name of the snapshot output
  * \param url the destination URL for the snapshot output
+ * \return 0 on success, else a negative LTTng error code
  */
-int lttng_add_snapshot_output(
+int add_snapshot_output(
   const std::string & session_name,
   const uint64_t max_size,
   const std::string & name,
@@ -38,8 +39,9 @@ int lttng_add_snapshot_output(
  * Record a snapshot session.
  *
  * \param session_name the session name
+ * \return 0 on success, else a negative LTTng error code
  */
-int lttng_record_snapshot(const std::string & session_name);
+int record_snapshot(const std::string & session_name);
 
 }  // namespace lttngpy
 
--- a/tracetools_launch/tracetools_launch/action.py
+++ b/tracetools_launch/tracetools_launch/action.py
@@ -113,7 +113,7 @@ class Trace(Action):
         self,
         *,
         session_name: SomeSubstitutionsType,
-        is_init_session: Union[bool, SomeSubstitutionsType] = False,
+        is_snapshot_session: Union[bool, SomeSubstitutionsType] = False,
         append_timestamp: Union[bool, SomeSubstitutionsType] = False,
         base_path: Optional[SomeSubstitutionsType] = None,
         append_trace: Union[bool, SomeSubstitutionsType] = False,
@@ -140,7 +140,7 @@ class Trace(Action):
         an empty string (through launch frontends).
 
         :param session_name: the name of the tracing session
-        :param is_init_session: whether the session is a init session
+        :param is_snapshot_session: whether the session is a snapshot session
         :param append_timestamp: whether to append timestamp to the session name
         :param base_path: the path to the base directory in which to create the session directory,
             or `None` for default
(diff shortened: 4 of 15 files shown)


tracetools_launch/tracetools_launch/action.py
```diff
@@ -113,10 +113,11 @@ def __init__(
         self,
         *,
         session_name: SomeSubstitutionsType,
+        is_init_session: Union[bool, SomeSubstitutionsType] = False,
         append_timestamp: Union[bool, SomeSubstitutionsType] = False,
         base_path: Optional[SomeSubstitutionsType] = None,
         append_trace: Union[bool, SomeSubstitutionsType] = False,
-        events_ust: Iterable[SomeSubstitutionsType] = names.DEFAULT_EVENTS_ROS,
+        events_ust: Iterable[SomeSubstitutionsType] = None,
```
christophebedard: I think this mechanism is good, but:

1. The documentation for `events_ust` should then mention that the default list of events used if `events_ust` is `None` (default) is `DEFAULT_EVENTS_ROS` if not `is_init_session` and otherwise it is `DEFAULT_INIT_EVENTS_ROS`.
1. The type annotation should be updated

```suggestion
        events_ust: Optional[Iterable[SomeSubstitutionsType]] = None,
```

tracetools_launch/tracetools_launch/action.py
```diff
@@ -113,10 +113,11 @@ def __init__(
         self,
         *,
         session_name: SomeSubstitutionsType,
+        is_init_session: Union[bool, SomeSubstitutionsType] = False,
```
christophebedard: I think we should call this `snapshot_session` session, since it could technically be used not _just_ for init events: you could use `is_snapshot_session=True` and `events_ust=DEFAULT_EVENTS_ROS` to do a snapshot with all events. In that case, the snapshot being used for init events is just 1 use case, even though it is probably the main use case (hence why we set `events_ustDEFAULT_INIT_EVENTS_ROS` if `snapshot_session`).

tracetools_launch/tracetools_launch/action.py
```diff
@@ -430,10 +446,15 @@ def execute(self, context: LaunchContext) -> List[Action]:
         )
         self._ld_preload_actions = self._get_ld_preload_actions(events_ust)
 
+        # Append '-init' to the session name if it is an init session
+        if is_init_session:
+            session_name += '-init'
```
christophebedard: Related to my other comment about the name of the `is_init_session` parameter and the overall snapshot feature vs the init events use-case, I would make this:

```suggestion
        # Append '-snapshot' to the session name if it is a snapshot session
        if is_init_session:
            session_name += '-snapshot'
```

However, maybe `'-snapshot'` should be added by `lttng_impl.setup()`? This way, it wouldn't have to remove `-snapshot` from the session name.

tracetools_trace/tracetools_trace/tools/names.py
```diff
@@ -104,6 +104,26 @@
...
+    tracepoints.rcl_service_init,
+    tracepoints.rmw_client_init,
+    tracepoints.rcl_client_init,
+    tracepoints.rcl_timer_init,
+    tracepoints.rclcpp_callback_register,
+    tracepoints.rcl_lifecycle_state_machine_init,
+    tracepoints.rclcpp_ipb_to_subscription,
+    tracepoints.rclcpp_buffer_to_ipb,
+    tracepoints.rclcpp_construct_ring_buffer,
+]
```
christophebedard: TODO for me: double-check this list.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -242,7 +243,11 @@ def setup(
     if not session_name:
         raise RuntimeError('empty session name')
     # Resolve full tracing directory path
-    full_path = os.path.join(base_path, session_name)
+    # For init and runtime sessions, remove the suffixes in the full path
+    if is_init_session:
+        full_path = os.path.join(base_path, session_name.removesuffix('-init'))
```
christophebedard: I think we discussed putting the snapshot session and the normal (runtime) session in the same directory (which would contain the 2 traces). Example:

* `~/.ros/tracing/${session_name}/${session_name}-snapshot/` for the snapshot (init events) trace
* `~/.ros/tracing/${session_name}/${session_name}/` for the normal (all events) trace

In this case, `base_path` would be `~/.ros/tracing/`. We would just need to do:

```suggestion
    # For init and runtime sessions, remove the suffixes in the full path
    if is_init_session:
        session_name_snapshot = session_name + '-snapshot'
        full_path = os.path.join(base_path, session_name, session_name_snapshot)
        session_name = session_name_snapshot
```

I think you _might_ have to create the `~/.ros/tracing/${session_name}/` directory, because I think LTTng only creates the trace directory (e.g., `~/.ros/tracing/${session_name}/${session_name}-snapshot/`) and expects the directory that contains it to exist.

christophebedard (commented):
Some initial comments!

tracetools_launch/tracetools_launch/action.py
ShravanDeva5327: Updated in https://github.com/ros2/ros2_tracing/pull/191/commits/9ca2cfd84186c1fb375f95ae5b0530d52d6e5ac4

tracetools_launch/tracetools_launch/action.py
ShravanDeva5327: Fixed in https://github.com/ros2/ros2_tracing/pull/191/commits/05ce13ca04b4ec4a36626e504b9548de68bde8dc

tracetools_trace/tracetools_trace/tools/names.py
ShravanDeva5327: Updated the list with new init events which were missing (https://github.com/ros2/ros2_tracing/pull/191/commits/5a8ff3e397057bfb08ed637aa4db79343ebf14c0)

tracetools_trace/tracetools_trace/tools/lttng_impl.py
ShravanDeva5327: Changed the destination directory of snapshot session to `~/.ros/tracing/{session-name}/snapshot` (https://github.com/ros2/ros2_tracing/pull/191/commits/9ca2cfd84186c1fb375f95ae5b0530d52d6e5ac4)

ShravanDeva5327: I am thinking modify `ros2 trace command` such that it will set the path of runtime traces to `~/.ros/tracing/{session-name}/runtime`, this will allow the users to open `~/.ros/tracing/{session-name}/` using babeltrace/tracecompass directly and also to distinguish between these two traces easily

tracetools_launch/tracetools_launch/action.py
christophebedard: I think the documentation about the default values for `events_ust` that I mentioned above is still missing.

lttngpy/src/lttngpy/snapshot.cpp
```diff
@@ -0,0 +1,37 @@
+#include <lttng/snapshot.h>
```
christophebedard, line 15: Missing copyright header

lttngpy/src/lttngpy/snapshot.cpp
```diff
@@ -0,0 +1,37 @@
...
+  lttng_snapshot_output_set_local_path(url.c_str(), output);
+
+  // Add snapshot output to the session.
+  lttng_snapshot_add_output(session_name.c_str(), output);
+
+  int ret = lttng_snapshot_record(session_name.c_str(), output, wait);
+  return ret;
+}
+
+}  // namespace lttngpy
```
christophebedard: To add a newline at the end of the file, as required by linters:

```suggestion
}  // namespace lttngpy

```

lttngpy/src/lttngpy/snapshot.hpp
```diff
@@ -0,0 +1,28 @@
+#ifndef LTTNGPY__SNAPSHOT_HPP_
```
christophebedard, line 15: Missing copyright header, and make sure to add an empty line between the copyright header and this `#ifndef` line

lttngpy/src/lttngpy/snapshot.hpp
```diff
@@ -0,0 +1,28 @@
...
+  const std::string & session_name,
+  const uint32_t id,
+  const uint64_t max_size,
+  const std::string & name,
+  const std::string & url,
+  const int wait);
+
+} // namespace lttngpy
+
+#endif  // LTTNGPY__SNAPSHOT_HPP_
```
christophebedard: Missing empty line at the end of the file to satisfy the linters

lttngpy/src/lttngpy/snapshot.cpp
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  // Set snapshot output attributes.
+  lttng_snapshot_output_set_id(id, output);
+  lttng_snapshot_output_set_size(max_size, output);
+  lttng_snapshot_output_set_name(name.c_str(), output);  
+  lttng_snapshot_output_set_local_path(url.c_str(), output);
+
+  // Add snapshot output to the session.
+  lttng_snapshot_add_output(session_name.c_str(), output);
```
christophebedard: I know it's annoying to write and ugly to read, but you should check the return code of each of these calls and return right away if there is an error, i.e., `if (0 != ret)` then `return ret;`. There probably won't be an error here, but if there is an error with any of these, we generally want to report the exact error as soon as possible, which means returning the error/return code.

lttngpy/src/lttngpy/snapshot.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+  int ret = lttng_snapshot_record(session_name.c_str(), output, wait);
```
christophebedard: Since the [`lttng_snapshot_record()` `wait` parameter is unused](https://github.com/lttng/lttng-tools/blob/a5f5b99b39c5f10acb26f35c89ce250690648267/include/lttng/snapshot.h#L165-L167), I think we could just put `1`/whatever here and not expose a `wait` parameter through `lttngpy.lttng_record_snapshot()`. `lttngpy` is a bit more flexible than the underlying lttng-ctl API, so we can simply add it when the `wait` parameter is actually used/needed.

tracetools_trace/tracetools_trace/tools/args.py
```diff
@@ -80,6 +80,9 @@ def _add_arguments_configure(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         '-a', '--append-trace', dest='append_trace', action='store_true',
         help='append to trace if it already exists, otherwise error out (default: %(default)s)')
+    parser.add_argument(
+        '-r', '--runtime', dest='is_runtime_session', action='store_true',
+        help='enable runtime tracing (default: %(default)s)')
```
christophebedard: So we discussed this in our meeting yesterday. I think it's fine, but I would rename this option to `--record-snapshot` or something like that. This is because this command already enables a normal "runtime" session, and what this option actually does is make it trigger a snapshot (with the given session name) _on top_ of creating a normal "runtime" session.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -479,6 +509,39 @@ def _create_session(
         raise RuntimeError(f"failed to create tracing session '{session_name}': {error}")
 
 
+def _create_session_snapshot(
+        *,
+        session_name: str,
+        full_path: Optional[str] = None,
+) -> None:
```
christophebedard:
```suggestion
def _create_session_snapshot(
    *,
    session_name: str,
    full_path: Optional[str] = None,
) -> None:
```

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -125,15 +134,20 @@ def init(
         display_list=display_list,
     )
 
+    if is_runtime_session:
+        session_name = session_name + '-runtime'
```
christophebedard: I see a lot of `session_name + '-runtime'` and `session_name.removesuffix('-runtime')`, in this file and in other files. Ideally, we'd want this logic (even though it's pretty trivial) in a single place for maintainability. However, it's hard to know how it could be improved without seeing the full implementation. Therefore, I will wait until then to take a closer look and potentially recommend a solution

lttngpy/src/lttngpy/snapshot.hpp
```diff
@@ -0,0 +1,28 @@
...
+ */
+int lttng_record_snapshot(
+  const std::string & session_name,
+  const uint32_t id,
+  const uint64_t max_size,
+  const std::string & name,
+  const std::string & url,
+  const int wait);
+
+} // namespace lttngpy
```
christophebedard:
```suggestion
}  // namespace lttngpy
```

christophebedard (commented):
Some comments and suggestions.

I will wait for the rest of the implementation to review the feature as a whole.

lttngpy/src/lttngpy/snapshot.cpp
christophebedard: As discussed, you can either put your name as the copyright holder, or you can put "Open Source Robotics Foundation, Inc." like here (make sure to put 2025 as the year): https://github.com/ros2/rclcpp/blob/b7e4aad091c6b334b9ad17a8f83007ea9901cc3b/rclcpp/include/rclcpp/rclcpp.hpp#L1-L13

tracetools_launch/tracetools_launch/action.py
ShravanDeva5327: Sorry, I missed it earlier.
Modified the doc string in https://github.com/ros2/ros2_tracing/pull/191/commits/cb868472c297c77198dcac004f3d1b42a17123f5

lttngpy/src/lttngpy/snapshot.cpp
ShravanDeva5327: Removed it in https://github.com/ros2/ros2_tracing/pull/191/commits/aaba91ecaef2c26b51f75bbcbdaedb9dccc5dd4a

lttngpy/src/lttngpy/snapshot.cpp
ShravanDeva5327: Added this in https://github.com/ros2/ros2_tracing/pull/191/commits/b76c5eda1ea4c26a6c9b530b6af0cddf4ba3ef06

ShravanDeva5327: I think we might get errors in `lttng_snapshot_add_output`, but we shouldn't get any errors in `lttng_snapshot_output_set_...` right (since these functions are just setting the values)?

christophebedard:
@ShravanDeva5327 as discussed earlier, we can strictly support the following in this PR:

1. `Trace(session_name='mysession', dual_session=True)` creates and starts a snapshot session and always sets up the directory structure and appends `'-snapshot'` to the session name
2. `ros2 trace -s mysession --dual-session` records the snapshot (started by the launch action) and starts a normal tracing session in the same directory as the snapshot session and with `'-runtime'` appended to the session name

We should also make sure that the help text for the `--dual-session` option for `ros2 trace` explains that it will record the snapshot session with the same name and start a normal tracing session.

tracetools_trace/tracetools_trace/tools/args.py
christophebedard: We discussed this today. Let's go with `--runtime` for this (i.e., `ros2 trace` recording the snapshot session and starting a normal tracing session with the same name). However, the help message should mention that, as mentioned in https://github.com/ros2/ros2_tracing/pull/191#issuecomment-3181049470.

lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
```diff
@@ -66,6 +67,22 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
...
+  m.def(
+    "lttng_create_session_snapshot",
+    &lttng_create_session_snapshot,
+    "Create snapshot session.",
+    py::kw_only(),
+    py::arg("session_name"),
+    py::arg("url"));
+  m.def(
+    "lttng_add_snapshot_output",
+    &lttngpy::lttng_add_snapshot_output,
```
christophebedard: I would rename this to `add_snapshot_output`, because the general convention here is that:

1. `lttng_*` functions exposed by `lttngpy` as-is keep the same name with the `lttng_` prefix, e.g., `lttng_destroy_session()`
2. Functions wrapped by a `lttngpy::` C++ function (which is then exposed by `lttngpy`) do _not_ keep the `lttng_` prefix, e.g., `destroy_all_sessions()`

```suggestion
    "add_snapshot_output",
    &lttngpy::add_snapshot_output,
```

lttngpy/src/lttngpy/snapshot.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <lttng/lttng-error.h>
+
+#include <string>
+
+#include "lttngpy/snapshot.hpp"
+
+namespace lttngpy
+{
+
+int lttng_add_snapshot_output(
```
christophebedard:
```suggestion
int add_snapshot_output(
```

lttngpy/src/lttngpy/snapshot.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Add an output object to the snapshot session.
+ *
+ * \param session_name the session name
+ * \param max_size the maximum size of the snapshot output in bytes
+ * \param name the name of the snapshot output
+ * \param url the destination URL for the snapshot output
+ */
+int lttng_add_snapshot_output(
```
christophebedard:
```suggestion
int add_snapshot_output(
```

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -479,6 +517,50 @@ def _create_session(
...
+        # destroy it and try again
+        destroy(session_name=session_name)
+        result = lttngpy.lttng_create_session_snapshot(
+            session_name=session_name,
+            url=None,
+        )
+    if result < 0:
+        error = lttngpy.lttng_strerror(result)
+        raise RuntimeError(f"failed to create tracing session '{session_name}': {error}")
+    result = lttngpy.lttng_add_snapshot_output(
```
christophebedard:
```suggestion
    result = lttngpy.add_snapshot_output(
```

lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
```diff
@@ -90,6 +107,12 @@ PYBIND11_MODULE(_lttngpy_pybind11, m) {
     "Stop tracing.",
     py::kw_only(),
     py::arg("session_name"));
+  m.def(
+    "lttng_record_snapshot",
+    &lttngpy::lttng_record_snapshot,
```
christophebedard: Similar to my other comment

```suggestion
    "record_snapshot",
    &lttngpy::record_snapshot,
```

lttngpy/src/lttngpy/snapshot.cpp
```diff
@@ -0,0 +1,62 @@
...
+  if (!try_or_cleanup(lttng_snapshot_output_set_size(max_size, output))) {return ret;}
+  if (!try_or_cleanup(lttng_snapshot_output_set_name(name.c_str(), output))) {return ret;}
+  if (!try_or_cleanup(lttng_snapshot_output_set_local_path(url.c_str(), output))) {return ret;}
+
+  // Add output object to the session.
+  ret = lttng_snapshot_add_output(session_name.c_str(), output);
+  return ret;
+}
+
+int lttng_record_snapshot(const std::string & session_name)
```
christophebedard:
```suggestion
int record_snapshot(const std::string & session_name)
```

lttngpy/src/lttngpy/snapshot.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+  int ret = lttng_snapshot_record(session_name.c_str(), NULL, 1);
+  return ret;
```
christophebedard:
```suggestion
  return lttng_snapshot_record(session_name.c_str(), NULL, 1);
```

Can you also add a comment to explain the `1` at the end? Maybe just "The last parameter is currently unused"

lttngpy/src/lttngpy/snapshot.hpp
```diff
@@ -0,0 +1,46 @@
...
+  const uint64_t max_size,
+  const std::string & name,
+  const std::string & url);
+
+/**
+ * Record a snapshot session.
+ *
+ * \param session_name the session name
+ */
+int lttng_record_snapshot(const std::string & session_name);
```
christophebedard:
```suggestion
int record_snapshot(const std::string & session_name);
```

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -427,6 +444,27 @@ def stop(
...
+) -> None:
+    """
+    Record a snapshot, and check for errors.
+
+    This must not be called if `lttngpy.is_available()` is `False`.
+    Raises RuntimeError on failure.
+
+    :param session_name: the name of the session
+    """
+    result = lttngpy.lttng_record_snapshot(session_name=session_name)
```
christophebedard:
```suggestion
    result = lttngpy.record_snapshot(session_name=session_name)
```

lttngpy/src/lttngpy/snapshot.hpp
```diff
@@ -0,0 +1,46 @@
...
+{
+
+/**
+ * Add an output object to the snapshot session.
+ *
+ * \param session_name the session name
+ * \param max_size the maximum size of the snapshot output in bytes
+ * \param name the name of the snapshot output
+ * \param url the destination URL for the snapshot output
+ */
```
christophebedard, line 31: This is missing a `\return`:

```suggestion
 * \return 0 on success, else a negative LTTng error code
 */
```

lttngpy/src/lttngpy/snapshot.hpp
```diff
@@ -0,0 +1,46 @@
...
+  const std::string & session_name,
+  const uint64_t max_size,
+  const std::string & name,
+  const std::string & url);
+
+/**
+ * Record a snapshot session.
+ *
+ * \param session_name the session name
+ */
```
christophebedard, line 43: This is missing a `\return`:

```suggestion
 * \return 0 on success, else a negative LTTng error code
 */
```

lttngpy/src/lttngpy/snapshot.cpp
```diff
@@ -0,0 +1,62 @@
...
+  const std::string & name,
+  const std::string & url)
+{
+  struct lttng_snapshot_output * output = lttng_snapshot_output_create();
+  if (nullptr == output) {
+    return -LTTNG_ERR_UNK;
+  }
+
+  int ret = 0;
+  auto try_or_cleanup = [&](int function_call) -> bool {
```
christophebedard: I would make this explicit for `ret` and `output`, just to be clear:

```suggestion
  auto try_or_cleanup = [&ret, output](int function_call) -> bool {
```

---

Also, this is good, but I wanted to highlight something here just for your information. This is quite a common pattern for doing "cleanup" when one call (out of multiple calls) to a C API fails (in this case: the lttng-ctl C API), since it is quite tedious, as you've noticed.

A lot of the ROS 2 code uses `rcpputils::make_scope_exit()`, which creates an object with a lambda that gets executed on scope exit, e.g., when the function `return`s due to a failure. If everything went well and there is no need to clean up, then we simply "cancel" it so that the lambda doesn't get executed. See for example this: https://github.com/ros2/rmw_fastrtps/blob/f4d412c7fdfdb15339c6de09b6ad667af11030b0/rmw_fastrtps_cpp/src/rmw_subscription.cpp#L98-L127. You can see here that the implementation is quite simple: it creates an object that simply calls the provided lambda in its destructor, which will therefore be called on scope exit https://github.com/ros2/rcpputils/blob/37a8801c18632d7d725ff1659248e8b12a7f2089/include/rcpputils/scope_exit.hpp#L31-L32.

The object is created with a lambda that does some cleanup (destroying the subscription object). Right before the function returns at the end, `cancel()` is called because at that point everything worked and we don't need to clean up.

I wouldn't use that here because I don't think it's worth making `lttngpy` depend on the `rcpputils` package strictly for `rcpputils::make_scope_exit()`. But this mechanism is quite common in C++, especially since it is easy to implement, although it will probably be added to the C++ STL eventually. Looks like it's experimentally available right now: https://en.cppreference.com/w/cpp/experimental/scope_exit.html

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -125,15 +150,35 @@ def init(
...
-    )
+    if is_runtime_session:
+        _display_snapshot_session_path(session_name=session_name, base_path=base_path)
+        base_path, full_runtime_session_path = _resolve_runtime_session_path(
+            session_name=session_name,
+            base_path=base_path,
+        )
+
+        snapshot_session_name = session_name + '-snapshot'
+        runtime_session_name = session_name + '-runtime'
```
christophebedard: There's a lot of `= session_name + '-runtime'` or `+= '-runtime'`. Can you create a constant for `'-runtime'`? Maybe in `tracetools_trace/tracetools_trace/tools/path.py`, and then you can use it here and in `lttng_impl.py`.

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -479,6 +517,50 @@ def _create_session(
...
+        raise RuntimeError(f"failed to create tracing session '{session_name}': {error}")
+    result = lttngpy.lttng_add_snapshot_output(
+        session_name=session_name,
+        max_size=max_size,
+        name=name,
+        url=full_path,
+    )
+    if result < 0:
+        error = lttngpy.lttng_strerror(result)
+        raise RuntimeError(f"failed to setup the tracing session '{session_name}': {error}")
```
christophebedard:
```suggestion
        raise RuntimeError(f"failed to add output for snapshot session '{session_name}': {error}")
```

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -479,6 +517,50 @@ def _create_session(
...
+        # Sessions may persist if there was an error previously, so if it already exists, just
+        # destroy it and try again
+        destroy(session_name=session_name)
+        result = lttngpy.lttng_create_session_snapshot(
+            session_name=session_name,
+            url=None,
+        )
+    if result < 0:
+        error = lttngpy.lttng_strerror(result)
+        raise RuntimeError(f"failed to create tracing session '{session_name}': {error}")
```
christophebedard:
```suggestion
        raise RuntimeError(f"failed to create snapshot session '{session_name}': {error}")
```

lttngpy/src/lttngpy/_lttngpy_pybind11.cpp
ShravanDeva5327: oh! changed the name of these functions now

lttngpy/src/lttngpy/snapshot.cpp
ShravanDeva5327: Thank you for sharing this!

ShravanDeva5327: Thank you for sharing this!

tracetools_trace/tracetools_trace/trace.py
ShravanDeva5327: Should I import `path.py` in `lttng_impl.py`? there is only one `+= '-runtime'`in this file

lttngpy/src/lttngpy/snapshot.cpp
christophebedard: we shouldn't get an error with `lttng_snapshot_output_set_*()`, but we could according to the API docs (f the inputs are invalid), so we should check

tracetools_launch/tracetools_launch/action.py
```diff
@@ -139,12 +140,15 @@ def __init__(
...
+        :param dual_session: whether to pre-configure a dual session
         :param append_timestamp: whether to append timestamp to the session name
         :param base_path: the path to the base directory in which to create the session directory,
             or `None` for default
         :param append_trace: whether to append to the trace directory if it already exists,
             otherwise an error is reported
-        :param events_ust: the list of ROS UST events to enable
+        :param events_ust: the list of ROS UST events to enable; if it's `None`, the default ROS
+            events are used for a normal session, and the default ROS initialization events are
+            used for a snapshot session; if it's an empty list, no UST events are enabled
```
christophebedard:
```suggestion
            used for the snapshot session in case of a dual session; if it's an empty list, no UST
            events are enabled
```

tracetools_launch/tracetools_launch/action.py
```diff
@@ -139,12 +140,15 @@ def __init__(
         an empty string (through launch frontends).
 
         :param session_name: the name of the tracing session
+        :param dual_session: whether to pre-configure a dual session
```
christophebedard:
```suggestion
        :param dual_session: whether to pre-configure a dual session: a snapshot session
            is configured here, with a normal tracing session to be configured later, while the
            application is running, e.g., through `ros2 trace --dual-session` with the same session
            name
```

tracetools_trace/tracetools_trace/tools/args.py
christophebedard: follow-up: we're going with `--dual-session`

tracetools_trace/tracetools_trace/trace.py
christophebedard: yes, it's still worth using the constants (`RUNTIME_SESSION_SUFFIX`, `SNAPSHOT_SESSION_SUFFIX`) in `lttng_impl.py` even if they're only used once

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -242,7 +246,12 @@ def setup(
     if not session_name:
         raise RuntimeError('empty session name')
     # Resolve full tracing directory path
-    full_path = os.path.join(base_path, session_name)
+    if snapshot_session and dual_session:
+        full_path = os.path.join(base_path, session_name.removesuffix('-snapshot'), 'snapshot')
+    elif dual_session:
+        full_path = os.path.join(base_path, session_name.removesuffix('-runtime'), 'runtime')
```
christophebedard: As mentioned in another thread, I would use `RUNTIME_SESSION_SUFFIX` and `SNAPSHOT_SESSION_SUFFIX` here

tracetools_launch/tracetools_launch/action.py
```diff
@@ -430,10 +448,16 @@ def execute(self, context: LaunchContext) -> List[Action]:
         )
         self._ld_preload_actions = self._get_ld_preload_actions(events_ust)
 
+        # Append '-snapshot' to the session name if pre-configuring a dual session
+        if dual_session:
+            session_name += '-snapshot'
```
christophebedard: This should use `SNAPSHOT_SESSION_SUFFIX`

tracetools_trace/tracetools_trace/tools/lttng_impl.py
```diff
@@ -427,6 +444,27 @@ def stop(
         raise RuntimeError(f"failed to stop tracing session '{session_name}': {error}")
 
 
+def record_snapshot(
+    *,
+    session_name: str,
+    **kwargs
```
christophebedard: Very minor, but I usually use trailing commas for the last item in lists (list of parameters in this case)

```suggestion
    **kwargs,
```

tracetools_launch/tracetools_launch/action.py
```diff
@@ -445,7 +469,8 @@ def setup() -> bool:
                 )
                 if self._trace_directory is None:
                     return False
-                self._logger.info(f'Writing tracing session to: {self._trace_directory}')
+                if not dual_session:
+                    self._logger.info(f'Writing tracing session to: {self._trace_directory}')
```
christophebedard: We should still print the trace directory if it's a dual session.

In the case of a dual session (snapshot session) here, the trace directory path returned by `lttng.lttng_init()` should be `~/.ros/tracing/$session_name/snapshot`. I think it's fine to print that here.

```suggestion
                self._logger.info(f'Writing tracing session to: {self._trace_directory}')
```

tracetools_trace/tracetools_trace/tools/args.py
```diff
@@ -89,18 +89,28 @@ def _add_arguments_default_session_name(parser: argparse.ArgumentParser) -> None
         help='the name of the tracing session (default: session-YYYYMMDDHHMMSS)')
 
 
+def _add_arguments_session_type(parser: argparse.ArgumentParser) -> None:
+    """Add session type argument to parser."""
+    parser.add_argument(
+        '-d', '--dual-session', dest='dual_session', action='store_true',
+        help='record snapshot of the pre-configured snapshot session '
+             'with the same name and start normal tracing session (default: %(default)s)')
```
christophebedard: I don't think I realized this earlier, but, since `--dual-session` is generally used for the dual session use-case and means different specific things depending on the `ros2 trace ...` command it's used with, this description (help text) is a bit incomplete. See https://github.com/ros2/ros2_tracing/issues/190#issuecomment-3166186813.

"Record a snapshot of the pre-configured snapshot session with the same name and start normal tracing session" only applies to `ros2 trace` (interactive) or `ros2 trace start` (non-interactive). For other commands (`ros2 trace pause/resume/stop`), the corresponding command (pause/resume/stop) applies to the normal (runtime) tracing session. So there are 2 groups.

We should probably change the help text based on the group that the given command belongs to. This can be done later. For now, the description should mention everything:

1. for the case where a snapshot session and a normal session are combined
2. when starting to trace (with a normal tracing session), this also triggers the corresponding snapshot
3. when pausing/resuming/stopping, this only affects the normal tracing session

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -261,7 +320,12 @@ def work() -> int:
                 interactive=False,
             )
         )
-    return _do_work_and_report_error(work, args.session_name, do_cleanup=True)
+    return _do_work_and_report_error(
+        work,
+        args.session_name,
+        args.dual_session,
+        do_cleanup=True
```
christophebedard:
```suggestion
        do_cleanup=True,
```

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -307,7 +377,12 @@ def resume(args: argparse.Namespace) -> int:
     :return: the return code (0 if successful, 1 otherwise)
     """
     def work() -> int:
-        lttng.lttng_start(session_name=args.session_name)
+        session_name = args.session_name
+        if args.dual_session:
+            snapshot_session_name = session_name + path.SNAPSHOT_SESSION_SUFFIX
+            lttng.lttng_record_snapshot(session_name=snapshot_session_name)
+            session_name += path.RUNTIME_SESSION_SUFFIX
+        lttng.lttng_start(session_name=session_name)
```
christophebedard: Interesting, I didn't think about triggering the snapshot when resuming the normal tracing session too! I think this would be worth a quick comment to point it out:

```suggestion
        session_name = args.session_name
        if args.dual_session:
            # Trigger snapshot to record any init events missed while paused
            snapshot_session_name = session_name + path.SNAPSHOT_SESSION_SUFFIX
            lttng.lttng_record_snapshot(session_name=snapshot_session_name)
            session_name += path.RUNTIME_SESSION_SUFFIX
        lttng.lttng_start(session_name=session_name)
```

christophebedard (commented):
Some last comments and suggestions!

christophebedard:
@ShravanDeva5327 it works great! Once the above (minor) suggestions and comments are addressed, I'll run CI!

tracetools_launch/tracetools_launch/action.py
ShravanDeva5327: I think we should say some thing like "Tracing session will be written to:" ?

ShravanDeva5327:
I've addressed these comments and I think we are good to go now!

tracetools_launch/tracetools_launch/action.py
```diff
@@ -445,7 +473,12 @@ def setup() -> bool:
...
+                else:
+                    self._logger.info(
+                        'Snapshot tracing session will be written to: '
+                        f'{self._trace_directory}')
```
christophebedard: Follow-up to https://github.com/ros2/ros2_tracing/pull/191#discussion_r2289195467 regarding the above change.

This is similar to passive vs active voice. It's usually better to use active voice for technical documentation, befcause it's simpler. See:

1. From a general writing perspective: https://www.grammarly.com/blog/sentences/active-vs-passive-voice/
2. From a technical documentation writing perspective: https://developers.google.com/tech-writing/one/active-voice

Active voice is closer to "_Writing_ snapshot tracing session to" while passive voice is closer to "Snapshot tracing session _will be written_ to"

```suggestion
                    self._logger.info(
                        f'Writing snapshot tracing session to: {self._trace_directory}')
```

tracetools_launch/tracetools_launch/action.py
christophebedard: See https://github.com/ros2/ros2_tracing/pull/191#discussion_r2291567646

christophebedard:
Seems like GitHub CI failed for unrelated reasons. Once my above suggestion is applied, it will be retriggered anyway

tracetools_launch/tracetools_launch/action.py
christophebedard: as discussed offline, we can just remove the condition and keep the original log

christophebedard:
Pulls: ros2/ros2_tracing#191
Gist: https://gist.githubusercontent.com/christophebedard/7a0fc1d8c9b34835f034c6fe4a3fd08e/raw/537ad47158708c7df0c80e9f223250a1590b8070/ros2.repos
BUILD args:  --packages-above-and-dependencies lttngpy ros2trace tracetools_launch tracetools_trace
TEST args:  --packages-above lttngpy ros2trace tracetools_launch tracetools_trace
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16765
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24670/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18696/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4118/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25192/)

christophebedard:
@ShravanDeva5327 as shown above, the `lttngpy` build fails on RHEL: https://ci.ros2.org/job/ci_linux-rhel/4118/console#console-section-26

```
rhel/ws/src/ShravanDeva5327/ros2_tracing/lttngpy/src/lttngpy/snapshot.cpp: In function ‘int lttngpy::add_snapshot_output(const string&, uint64_t, const string&, const string&)’:
13:45:55 /home/jenkins-agent/workspace/ci_linux-rhel/ws/src/ShravanDeva5327/ros2_tracing/lttngpy/src/lttngpy/snapshot.cpp:49:23: error: ‘lttng_snapshot_output_set_local_path’ was not declared in this scope; did you mean ‘lttng_snapshot_output_set_data_url’?
13:45:55    49 |   if (!try_or_cleanup(lttng_snapshot_output_set_local_path(url.c_str(), output))) {return ret;}
13:45:55       |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
13:45:55       |                       lttng_snapshot_output_set_data_url
```

RHEL has an older version of the LTTng packages, specifically liblttng-ctl here. You can find it right above that, in the build logs/output:

```
13:45:55 --   Found lttng-ctl, version 2.12.11
```

Looks like `lttng_snapshot_output_set_local_path()` indeed doesn't exist in 2.12: https://github.com/lttng/lttng-tools/blob/stable-2.12/include/lttng/snapshot.h. But `lttng_snapshot_output_set_data_url()` exists for 2.12+ (Ubuntu currently has 2.13). Comparing both, it seems like they're almost the same thing: https://github.com/lttng/lttng-tools/blob/810703b69692d0d9ffc8f0663f94da2b3d61af8c/src/lib/lttng-ctl/snapshot.c#L361-L413. The difference is that `lttng_snapshot_output_set_local_path()` expects the input to be a local file path, so it may or may not start with `file://`: https://github.com/lttng/lttng-tools/blob/810703b69692d0d9ffc8f0663f94da2b3d61af8c/include/lttng/snapshot.h#L69-L77. We indeed seem to be providing a path/URL without `file://`.

I think calling `lttng_snapshot_output_set_data_url()` with `("file://" + url).c_str()` would be equivalent to `lttng_snapshot_output_set_local_path()` with `url.c_str()` and would work for 2.12+. Can you make this change?

ShravanDeva5327:
> I think calling lttng_snapshot_output_set_data_url() with ("file://" + url).c_str() would be equivalent to lttng_snapshot_output_set_local_path() with url.c_str() and would work for 2.12+. Can you make this change?

Yes, I'll try this once

ShravanDeva5327:
It looks like the `lttng_snapshot_output_set_local_path()` function is [setting the attribute `ctrl_url` of the output object](https://github.com/lttng/lttng-tools/blob/810703b69692d0d9ffc8f0663f94da2b3d61af8c/src/lib/lttng-ctl/snapshot.c#L404) just [like the `lttng_snapshot_output_set_ctrl_url()`](https://github.com/lttng/lttng-tools/blob/810703b69692d0d9ffc8f0663f94da2b3d61af8c/src/lib/lttng-ctl/snapshot.c#L351) function and not like the `lttng_snapshot_output_set_data_url()` function. Am I getting this wrong?

christophebedard:
Oh! No, you're right! Then the equivalent would be `lttng_snapshot_output_set_ctrl_url()` with `("file://" + url).c_str()`.

christophebedard:
Pulls: ros2/ros2_tracing#191
Gist: https://gist.githubusercontent.com/christophebedard/2c4ae7f6ce2ca4e8ff657b4274a692a8/raw/537ad47158708c7df0c80e9f223250a1590b8070/ros2.repos
BUILD args:  --packages-above-and-dependencies lttngpy ros2trace tracetools_launch tracetools_trace
TEST args:  --packages-above lttngpy ros2trace tracetools_launch tracetools_trace
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16780
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24687/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18711/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4134/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25206/)

christophebedard (approved):
Alright, CI looks good! Thanks for the PR and for iteraring with me!

ros-discourse:

ros-discourse:

PR #195, opened 2025-08-23
Allow creating snapshot sessions

ShravanDeva5327:
## Description

These changes will allow users to start tracing sessions in [snapshot mode](https://lttng.org/docs/v2.13/#doc-taking-a-snapshot) using both `Trace` action and the `ros2 trace` commands and to stop/pause/resume/record snapshot of the tracing session using `ros2 trace` command.

### Configuring/starting the tracing session
```python
Trace(
    session_name='my-snapshot-session',
    snapshot_mode=True,
),
```
or
```bash
# interactive tracing
ros2 trace -s my-snapshot-session --snapshot-mode
```
or
```bash
# non-interactive tracing
ros2 trace start my-snapshot-session --snapshot-mode
```

### Controlling the tracing session
```bash
ros2 trace pause my-snapshot-session
ros2 trace resume my-snapshot-session
ros2 trace stop my-snapshot-session 

ros2 trace record_snapshot my-snapshot-session
```

Note: The `--snapshot-mode` option is not required. Using `record_snapshot` with a non-snapshot session will result in an error.

Yes.

### DIFF
--- a/lttngpy/src/lttngpy/session.cpp
+++ b/lttngpy/src/lttngpy/session.cpp
@@ -42,6 +42,26 @@ std::variant<int, std::set<std::string>> get_session_names()
   return session_names;
 }
 
+std::variant<int, std::set<std::string>> get_snapshot_session_names()
+{
+  struct lttng_session * sessions = nullptr;
+  int ret = lttng_list_sessions(&sessions);
+  if (0 > ret) {
+    std::free(sessions);
+    return ret;
+  }
+
+  std::set<std::string> snapshot_session_names = {};
+  const int num_sessions = ret;
+  for (int i = 0; i < num_sessions; i++) {
+    if (sessions[i].snapshot_mode) {
+      snapshot_session_names.insert(sessions[i].name);
+    }
+  }
+  std::free(sessions);
+  return snapshot_session_names;
+}
+
 int destroy_all_sessions()
 {
   const auto & session_names_opt = lttngpy::get_session_names();
--- a/lttngpy/src/lttngpy/session.cpp
+++ b/lttngpy/src/lttngpy/session.cpp
@@ -24,7 +24,7 @@
 namespace lttngpy
 {
 
-std::variant<int, std::set<std::string>> get_session_names()
+std::variant<int, std::set<std::string>> get_session_names(bool snapshot_mode)
 {
   struct lttng_session * sessions = nullptr;
   int ret = lttng_list_sessions(&sessions);
@@ -36,30 +36,13 @@ std::variant<int, std::set<std::string>> get_session_names()
   std::set<std::string> session_names = {};
   const int num_sessions = ret;
   for (int i = 0; i < num_sessions; i++) {
+    if (snapshot_mode && !sessions[i].snapshot_mode) {
+      continue;
+    }
     session_names.insert(sessions[i].name);
   }
   std::free(sessions);
   return session_names;
-}
-
-std::variant<int, std::set<std::string>> get_snapshot_session_names()
-{
-  struct lttng_session * sessions = nullptr;
-  int ret = lttng_list_sessions(&sessions);
-  if (0 > ret) {
-    std::free(sessions);
-    return ret;
-  }
-
-  std::set<std::string> snapshot_session_names = {};
-  const int num_sessions = ret;
-  for (int i = 0; i < num_sessions; i++) {
-    if (sessions[i].snapshot_mode) {
-      snapshot_session_names.insert(sessions[i].name);
-    }
-  }
-  std::free(sessions);
-  return snapshot_session_names;
 }
 
 int destroy_all_sessions()
--- a/lttngpy/src/lttngpy/session.hpp
+++ b/lttngpy/src/lttngpy/session.hpp
@@ -25,16 +25,10 @@ namespace lttngpy
 /**
  * Get the currently-existing session names.
  *
+ * \param snapshot_mode If true, only snapshot session names are returned
  * \return the set of session names, else a negative LTTng error code
  */
-std::variant<int, std::set<std::string>> get_session_names();
-
-/**
- * Get the currently-existing snapshot session names.
- *
- * \return the set of snapshot session names, else a negative LTTng error code
- */
-std::variant<int, std::set<std::string>> get_snapshot_session_names();
+std::variant<int, std::set<std::string>> get_session_names(bool snapshot_mode = false);
 
 /**
  * Destroy all sessions.
--- a/lttngpy/src/lttngpy/session.hpp
+++ b/lttngpy/src/lttngpy/session.hpp
@@ -30,6 +30,13 @@ namespace lttngpy
 std::variant<int, std::set<std::string>> get_session_names();
 
 /**
+ * Get the currently-existing snapshot session names.
+ *
+ * \return the set of snapshot session names, else a negative LTTng error code
+ */
+std::variant<int, std::set<std::string>> get_snapshot_session_names();
+
+/**
  * Destroy all sessions.
  *
  * Tries to destroy all sessions, and reports an error if any session destruction was unsuccessful.
--- a/lttngpy/test/test_session.py
+++ b/lttngpy/test/test_session.py
@@ -59,6 +59,42 @@ class TestSession(unittest.TestCase):
 
         shutil.rmtree(tmpdir)
 
+    def test_snapshot_session_list_create_start_stop_destroy(self):
+        session_name = 'test_snapshot_session_list_create_start_stop_destroy'
+        tmpdir = self.create_test_tmpdir(session_name)
+
+        self.assertSetEqual(set(), lttngpy.get_session_names(snapshot_mode=True))
+        self.assertEqual(
+            0,
+            lttngpy.lttng_create_session_snapshot(session_name=session_name, url=tmpdir),
+        )
+        self.assertSetEqual({session_name}, lttngpy.get_session_names(snapshot_mode=True))
+        self.assertEqual(
+            0,
+            lttngpy.enable_channel(
+                session_name=session_name,
+                domain_type=lttngpy.LTTNG_DOMAIN_UST,
+                buffer_type=lttngpy.LTTNG_BUFFER_PER_UID,
+                channel_name='dummy_channel',
+                overwrite=None,
+                subbuf_size=None,
+                num_subbuf=None,
+                switch_timer_interval=None,
+                read_timer_interval=None,
+                output=None,
+            ),
+        )
+        self.assertEqual(0, lttngpy.lttng_start_tracing(session_name=session_name))
+        self.assertEqual(0, lttngpy.lttng_stop_tracing(session_name=session_name))
+        self.assertEqual(0, lttngpy.lttng_destroy_session(session_name=session_name))
+        self.assertSetEqual(set(), lttngpy.get_session_names(snapshot_mode=True))
+
+        self.assertEqual(0, lttngpy.lttng_create_session(session_name=session_name, url=tmpdir))
+        self.assertEqual(0, lttngpy.destroy_all_sessions())
+        self.assertSetEqual(set(), lttngpy.get_session_names(snapshot_mode=True))
+
+        shutil.rmtree(tmpdir)
+
     def test_error(self):
         session_name = 'test_error'
         self.assertSetEqual(set(), lttngpy.get_session_names())
--- a/lttngpy/test/test_session.py
+++ b/lttngpy/test/test_session.py
@@ -89,7 +89,10 @@ class TestSession(unittest.TestCase):
         self.assertEqual(0, lttngpy.lttng_destroy_session(session_name=session_name))
         self.assertSetEqual(set(), lttngpy.get_session_names(snapshot_mode=True))
 
-        self.assertEqual(0, lttngpy.lttng_create_session(session_name=session_name, url=tmpdir))
+        self.assertEqual(
+            0,
+            lttngpy.lttng_create_session_snapshot(session_name=session_name, url=tmpdir),
+        )
         self.assertEqual(0, lttngpy.destroy_all_sessions())
         self.assertSetEqual(set(), lttngpy.get_session_names(snapshot_mode=True))
 
--- a/ros2trace/ros2trace/verb/record_snapshot.py
+++ b/ros2trace/ros2trace/verb/record_snapshot.py
@@ -18,10 +18,10 @@ from tracetools_trace.trace import record_snapshot
 
 
 class RecordSnapshotVerb(VerbExtension):
-    """Record a snapshot of the current tracing session."""
+    """Take a snapshot of a snapshot session."""
 
-    def add_arguments(self, parser, cli_name):
+    def add_arguments(self, parser, cli_name) -> None:
         args.add_arguments_noninteractive_control(parser)
 
-    def main(self, *, args):
+    def main(self, *, args) -> int:
         return record_snapshot(args)
--- a/ros2trace/ros2trace/verb/record_snapshot.py
+++ b/ros2trace/ros2trace/verb/record_snapshot.py
@@ -1,0 +1,27 @@
+# Copyright 2025 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2trace.verb import VerbExtension
+from tracetools_trace.tools import args
+from tracetools_trace.trace import record_snapshot
+
+
+class RecordSnapshotVerb(VerbExtension):
+    """Record a snapshot of the current tracing session."""
+
+    def add_arguments(self, parser, cli_name):
+        args.add_arguments_noninteractive_control(parser)
+
+    def main(self, *, args):
+        return record_snapshot(args)
--- a/test_ros2trace/test/test_ros2trace/test_trace.py
+++ b/test_ros2trace/test/test_ros2trace/test_trace.py
@@ -99,11 +99,30 @@ class TestROS2TraceCLI(unittest.TestCase):
             f"tracing session '{session_name}' does not exist",
         )
 
+    def assertSnapshotTracingSessionExist(self, session_name: str) -> None:
+        self.assertTrue(
+            lttngpy.is_lttng_session_daemon_alive(),
+            f"tracing session '{session_name}' does not exist because there is no daemon",
+        )
+        session_names = lttngpy.get_snapshot_session_names()
+        self.assertIn(
+            session_name,
+            session_names,
+            f"tracing session '{session_name}' does not exist",
+        )
+
     def assertTracingSessionNotExist(self, session_name: str) -> None:
         # If there is no session daemon, then there are no tracing sessions
         if not lttngpy.is_lttng_session_daemon_alive():
             return
         session_names = lttngpy.get_session_names()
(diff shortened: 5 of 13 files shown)


ros2trace/ros2trace/verb/record_snapshot.py
```diff
@@ -0,0 +1,27 @@
...
+
+from ros2trace.verb import VerbExtension
+from tracetools_trace.tools import args
+from tracetools_trace.trace import record_snapshot
+
+
+class RecordSnapshotVerb(VerbExtension):
+    """Record a snapshot of the current tracing session."""
+
+    def add_arguments(self, parser, cli_name):
```
christophebedard:
```suggestion
    def add_arguments(self, parser, cli_name) -> None:
```

ros2trace/ros2trace/verb/record_snapshot.py
```diff
@@ -0,0 +1,27 @@
...
+        args.add_arguments_noninteractive_control(parser)
+
+    def main(self, *, args):
```
christophebedard:
```suggestion
    def main(self, *, args) -> int:
```

ros2trace/ros2trace/verb/record_snapshot.py
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from ros2trace.verb import VerbExtension
+from tracetools_trace.tools import args
+from tracetools_trace.trace import record_snapshot
+
+
+class RecordSnapshotVerb(VerbExtension):
+    """Record a snapshot of the current tracing session."""
```
christophebedard: This is a standalone command: the user has to explicitly provide a (snapshot) session name, so there is no "current tracing session." This should probably just say "... of a snapshot session"

Also, I think we may not want to refer to this as "_Record(ing)_ a snapshot." The corresponding LTTng command is `lttng snapshot record`, but LTTng doesn't really refer to this action as "_recording_ a snapshot" in the `lttng snapshot` documentation. It instead says "_take_ a snapshot" https://lttng.org/man/1/lttng-snapshot/v2.13/

> **lttng-snapshot** — Take a snapshot of an LTTng recording session
> ...
> Take a recording session snapshot:
> ...
> The `lttng snapshot` command can take a snapshot of, add or remove a snapshot output, and show the snapshot output of:

So I think this should be:

```suggestion
    """Take a snapshot of a snapshot session."""
```

ros2trace/ros2trace/verb/record_snapshot.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def add_arguments(self, parser, cli_name):
+        args.add_arguments_noninteractive_control(parser)
```
christophebedard, line 24: `ros2 trace {pause,resume,stop,record_snapshot} --help` includes `--snapshot-session`, which isn't valid, as you pointed out in the PR description. Maybe `add_arguments_noninteractive_control()` should have a `snapshot: bool` parameter to optionally include `--snapshot-session`?

tracetools_launch/tracetools_launch/action.py
```diff
@@ -113,6 +113,7 @@ def __init__(
         self,
         *,
         session_name: SomeSubstitutionsType,
+        snapshot_session: Union[bool, SomeSubstitutionsType] = False,
```
christophebedard: What do you think about calling this just `snapshot` instead of `snapshot_session` to keep it simple? It is already sort of implied that this concerns the session, since the `Trace` action configures a tracing session. Same thing for the name for launch XML/YAML frontends, and `--snapshot-session` --> `--snapshot` with `ros2 trace ...`. And the underlying `snapshot_session` option in `lttng_impl.setup()` could be simplified too.

It does look a bit weird if we have `snapshot` without `_session`, and `dual_session` (with `_session`), but "dual" is more generic, so `_session` makes more sense there.

What do you think?

christophebedard (changes requested):
Some comments and suggestions.

christophebedard:
I think this PR should include tests. We can start with testing the `Trace` action on its own and test the `ros2 trace ...` commands on their own too in this PR:

1. `Trace` action tests in `test_tracetools_launch/test/test_tracetools_launch/test_trace_action.py`
    1. Update XML and YAML tests (`test_action_frontend_xml`, `test_action_frontend_yaml`) to include the snapshot and dual session parameters. You can just set both to `false` in the test and then check that they were parsed into `False`.
    2. Similarly, update the substitutions test (`test_action_substitutions`, `test_action_substitutions_frontend_xml`, `test_action_substitutions_frontend_yaml`) to test substiutions for the snapshot and dual session parameters. You can make both `False`/`"false"` here too and then check that the substitutions were performed correctly.
    3. Add a test similar to `test_action` that simply enables the dual session option
    4. Add a test similar to `test_action` that simply enables the snapshot session option
2. `ros2 trace ...` tests in `test_ros2trace/test/test_ros2trace/test_trace.py`
    1. Add test for `ros2 trace start -s ... --snapshot`, `ros2 trace record_snapshot ...`, and `ros2 trace stop ...`, similar to `test_start_pause_resume_stop`
    2. Similar to the above test, you could add test using interactive tracing like `test_default_tracing` does: `ros2 trace --snapshot` with `ros2 trace record_snapshot ...` in between interactive start/stop. However, testing interactive tracing in tests might be problematic (see #166), so I would hold off on this for now

Then later we can add a test for `Trace` with dual session + `ros2 trace ... --dual-session`. I think that test will probably go in `test_ros2trace` too, since `test_ros2trace` already uses `launch` in its tests and already (indirectly) depends on `tracetools_launch`, but we can talk about it after this PR is done.

christophebedard:
By the way, there seems to be a problem with taking multiple snapshots with the same snapshot session:

1. Start launch file with snapshot session and some nodes
    ```console
    $ ros2 launch tracetools_launch snapshot_example.launch.py  # like example.launch.py but with snapshot enabled
    ```
2. Check that the snapshot session exists
    ```console
    $ lttng list
    Available recording sessions:
      1) my-tracing-session [active snapshot]


    Use lttng list <session_name> for more details
    ```
3. Take snapshot
    ```console
    $ ros2 trace record_snapshot my-tracing-session
    ```
4. Check trace data: the snapshot output is named `snapshot-20250823-133144-0` because it was taken at ~1:31 pm (local time) on Aug 23rd
    ```console
    $ babeltrace2 ~/.ros/tracing/my-tracing-session/snapshot-20250823-133144-0/
    ...
    ```
    - The trace data seems to include events starting from when the launch file was launched until when the snapshot was taken, which makes sense
5. Take another snapshot
    ```console
    $ ros2 trace record_snapshot my-tracing-session
    ```
6. Check trace data: the snapshot output is named `snapshot-20250823-133210-1` because it was taken at ~1:32 pm
    ```console
    $ babeltrace2 ~/.ros/tracing/my-tracing-session/snapshot-20250823-133210-1/
    ....
    ```
    - **Expected**: it contains data up until the moment this (_second_) snapshot was taken (not sure if it should include data from before the _first_ snapshot was taken, but I would expect it to contain newer data)
    - **Actual**: it contains the same exact data as the previous snapshot; it doesn't contain newer data
7. Stop launch with Ctrl+C
8. Check that no (snapshot) session exists
    ```console
    $ lttng list
    Currently no available recording session
    ```

Do you know why this happens?

ros2trace/ros2trace/verb/record_snapshot.py
ShravanDeva5327: Or may be I can add `--dual-session` into `add_arguments_noninteractive_control()` and `--snapshot-session` into `add_arguments_noninteractive_configure()`  and remove the [`_add_arguments_session_type ()`](https://github.com/ros2/ros2_tracing/blob/rolling/tracetools_trace/tracetools_trace/tools/args.py#L92) function?

tracetools_launch/tracetools_launch/action.py
ShravanDeva5327: I thing `snapshot` makes it a bit unclear, how about `snapshot_mode`?

ros2trace/ros2trace/verb/record_snapshot.py
ShravanDeva5327: got it! I'll change this

ShravanDeva5327:
> I think this PR should include tests. We can start with testing the `Trace` action on its own and test the `ros2 trace ...` commands on their own too in this PR:

I've started modifying the tests for `Trace` action. I tried adding a test similar to `test_append_timestamp` but not able to figure out how to verify that the session created is a snapshot session, do you have some suggestions on how to do this or it is sufficient if we assert that there are no errors in launching?

ShravanDeva5327:
> * **Expected**: it contains data up until the moment this (_second_) snapshot was taken (not sure if it should include data from before the _first_ snapshot was taken, but I would expect it to contain newer data)
> * **Actual**: it contains the same exact data as the previous snapshot; it doesn't contain newer data

I am not sure why this is happening, I would expect something like: the second trace contains data from the beginning of launch file to the time when second snapshot was triggered (since we are not clearing the subbuffers after taking the first snapshot).

There is an `lttng clear` command (See [this](https://lttng.org/man/1/lttng-clear/v2.13/)) that clears the contents of recording buffers, I didn't use it earlier because the documentation also state that this command clears the contents of all their local and streamed trace data.

> The `lttng clear `command clears one or more recording sessions, that is, it deletes the contents of their recording buffers and of all their local and streamed trace data.

I tested this locally just now with a snapshot session and it didn't clear/delete the trace data already written to the disk. I tried taking another snapshot after clearing the session and as expected this second snapshot has the trace data of events that happened after clearing the session and before triggering the new snapshot!

Do you want me to implement something like clearing the tracing session immediately after taking a snapshot?

ShravanDeva5327:
```c
/*
 * For snapshot sessions (flight recorder), only the buffer content is cleared.
 * Prior snapshots are individually recorded to disk, and are therefore
 * untouched by this "clear" command.
 */
```

this documentation comment in [`clear.h`](https://github.com/lttng/lttng-tools/blob/stable-2.13/include/lttng/clear.h#L29-L31)confirms that for snapshot sessions, the prior snapshots are not cleared/deleted.

tracetools_launch/tracetools_launch/action.py
christophebedard: `snapshot_mode`/`snapshot-mode`/`--snapshot-mode` is good!

ros2trace/ros2trace/verb/record_snapshot.py
christophebedard: I can't really tell or figure out how that would look like right now, but if it makes sense (with the function names) and doesn't duplicate options, yeah it should be fine

christophebedard:
> not able to figure out how to verify that the session created is a snapshot session, do you have some suggestions on how to do this or it is sufficient if we assert that there are no errors in launching?

There's an `lttngpy` function that lists session names, which is used in tests: https://github.com/ros2/ros2_tracing/blob/b0a3aef873da37d41693f88573225e14b3854bdc/test_ros2trace/test/test_ros2trace/test_trace.py#L95. You could add a boolean parameter (default to `False` to keep the current behaviour) called `snapshot` that would make it return snapshot session names only.

I think you can figure out if a session is a snapshot session with the `lttng_session` struct for a given session: https://github.com/ros2/ros2_tracing/blob/b0a3aef873da37d41693f88573225e14b3854bdc/lttngpy/src/lttngpy/session.cpp#L29

christophebedard:
> Do you want me to implement something like clearing the tracing session immediately after taking a snapshot?

> this documentation comment in [`clear.h`](https://github.com/lttng/lttng-tools/blob/stable-2.13/include/lttng/clear.h#L29-L31)confirms that for snapshot sessions, the prior snapshots are not cleared/deleted.

Yeah, that seems to be what we should do then.

christophebedard:
> > Do you want me to implement something like clearing the tracing session immediately after taking a snapshot?
> 
> > this documentation comment in [`clear.h`](https://github.com/lttng/lttng-tools/blob/stable-2.13/include/lttng/clear.h#L29-L31)confirms that for snapshot sessions, the prior snapshots are not cleared/deleted.
> 
> Yeah, that seems to be what we should do then.

As discussed today, we should also call `lttng_clear_handle_wait_for_completion()` (and `lttng_clear_handle_get_result()`) after `lttng_clear_session()`. And we can just call `lttng_clear_session()` right after triggering the snapshot recording.

tracetools_trace/tracetools_trace/trace.py
```diff
@@ -388,5 +393,24 @@ def work() -> int:
...
+
+    :param args: the arguments parsed using
+        `tracetools_trace.tools.args.add_arguments_noninteractive_control`
+    :return: the return code (0 if successful, 1 otherwise)
+    """
+    def work() -> int:
+        session_name = args.session_name
+        if args.dual_session:
+            session_name += path.RUNTIME_SESSION_SUFFIX
+        lttng.lttng_record_snapshot(session_name=session_name)
```
christophebedard: If a user does `ros2 trace record_snapshot session-name --dual-session`, they're intending to take a snapshot of the snapshot (not runtime) session, right? So this should be:

```suggestion
        session_name = args.session_name
        if args.dual_session:
            session_name += path.SNAPSHOT_SESSION_SUFFIX
        lttng.lttng_record_snapshot(session_name=session_name)
```

ShravanDeva5327: Yeah, I made a mistake. I'll fix this!

ShravanDeva5327:
I am planning to open another PR that will add tests for dual_session specifically, and not mix up these two. Will it be fine?

christophebedard:
> I am planning to open another PR that will add tests for dual_session specifically, and not mix up these two. Will it be fine?

Yeah, that's fine!

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -662,6 +681,76 @@ def test_start_pause_resume_stop(self) -> None:
...
+
+        # Subbuffers should be cleared after the last record_snapshot
+        ret = self.run_trace_subcommand(['record_snapshot', session_name])
+        self.assertEqual(0, ret)
+        self.assertSnapshotTracingSessionExist(session_name)
+        expected_trace_data = []
+        new_num_events = self.assertTraceContains(
+            trace_dir,
+            expected_field_value=expected_trace_data
+            )
```
christophebedard:
```suggestion
        new_num_events = self.assertTraceContains(
            trace_dir,
            expected_field_value=expected_trace_data,
        )
```

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -662,6 +681,76 @@ def test_start_pause_resume_stop(self) -> None:
...
+
+        # Stop tracing and check that session does not exist
+        ret = self.run_trace_subcommand(['stop', session_name])
+        self.assertEqual(0, ret)
+        self.assertSnapshotTracingSessionNotExist(session_name)
+
+        # Stopping tracing again should give an error
+        ret = self.run_trace_subcommand(['stop', session_name])
+        self.assertEqual(1, ret)
+        self.assertSnapshotTracingSessionNotExist(session_name)
```
christophebedard: Can you call `record_snapshot` after this and check that it properly returns an error/1 since the session (name) doesn't exist?

lttngpy/src/lttngpy/session.cpp
```diff
@@ -42,6 +42,26 @@ std::variant<int, std::set<std::string>> get_session_names()
...
+  std::set<std::string> snapshot_session_names = {};
+  const int num_sessions = ret;
+  for (int i = 0; i < num_sessions; i++) {
+    if (sessions[i].snapshot_mode) {
+      snapshot_session_names.insert(sessions[i].name);
+    }
+  }
+  std::free(sessions);
+  return snapshot_session_names;
+}
```
christophebedard: This is very similar to `get_session_names()`; they should be merged. You can add a parameter like `bool snapshot` (`get_session_names(bool snapshot)` above in the definition, and set `get_session_names(bool snapshot = false)` for the declaration in the header). Then you add a condition around `session_names.insert(sessions[i].name);` to only insert if `!snapshot` (i.e., the user didn't specifically request only snapshot session names) or if `snapshot` and `sessions[i].snapshot_mode` (i.e., the user requested only snapshot session names and this is a snapshot session).

You will also have to add this parameter and its default value (`False`) to the pybind11 wrapper (`_lttngpy_pybind11.cpp`), see: https://pybind11.readthedocs.io/en/stable/basics.html#default-args.

Similarly, you can then add a `snapshot = False` parameter to `assertTracingSession[Not]Exist()` and remove `assertSnapshotTracingSession[Not]Exist()`.

christophebedard (changes requested):
This is looking good! I have some minor comments/requests.

lttngpy/src/lttngpy/session.cpp
ShravanDeva5327: I tried doing this earlier but didn't add default value in `_lttngpy_pybind11.cpp`, and hence it was complaining when no argument is being passed. I'll change this now!

lttngpy/src/lttngpy/session.hpp
```diff
@@ -25,9 +25,10 @@ namespace lttngpy
 /**
  * Get the currently-existing session names.
  *
+ * \param snapshot_mode If true, only snapshot session names are returned
  * \return the set of session names, else a negative LTTng error code
  */
-std::variant<int, std::set<std::string>> get_session_names();
+std::variant<int, std::set<std::string>> get_session_names(bool snapshot_mode = false);
```
christophebedard, line 31: One last thing; sorry for not mentioning it earlier. There is a simple test for `get_session_names()` in `lttngpy`: https://github.com/ros2/ros2_tracing/blob/e0f37b6ff032b82e9255d84d3b9769db28650534/lttngpy/test/test_session.py#L29. Can you add another test that does the same thing but with a snapshot session?

ShravanDeva5327: done!

lttngpy/test/test_session.py
```diff
@@ -59,6 +59,42 @@ def test_session_list_create_start_stop_destroy(self):
...
+                read_timer_interval=None,
+                output=None,
+            ),
+        )
+        self.assertEqual(0, lttngpy.lttng_start_tracing(session_name=session_name))
+        self.assertEqual(0, lttngpy.lttng_stop_tracing(session_name=session_name))
+        self.assertEqual(0, lttngpy.lttng_destroy_session(session_name=session_name))
+        self.assertSetEqual(set(), lttngpy.get_session_names(snapshot_mode=True))
+
+        self.assertEqual(0, lttngpy.lttng_create_session(session_name=session_name, url=tmpdir))
```
christophebedard: (double check the line length)

```suggestion
        self.assertEqual(0, lttngpy.lttng_create_session_snapshot(session_name=session_name, url=tmpdir))
```

christophebedard (approved):
Looks good! Thanks for adding tests.

Could you just update the PR description to reflect the option name change to `--snapshot-mode`/`snapshot_mode`?

christophebedard:
Pulls: ros2/ros2_tracing#195
Gist: https://gist.githubusercontent.com/christophebedard/f09dfa06898f2aa43e1189d2a6d3feb3/raw/f430445a2b08d12a7cafd9bf87d96cec7abff1d6/ros2.repos
BUILD args:  --packages-above-and-dependencies lttngpy ros2trace test_ros2trace test_tracetools_launch tracetools_launch tracetools_trace
TEST args:  --packages-above lttngpy ros2trace test_ros2trace test_tracetools_launch tracetools_launch tracetools_trace
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/16846
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/24757/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/18785/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/4209/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25273/)

ShravanDeva5327:
> Could you just update the PR description to reflect the option name change to --snapshot-mode/snapshot_mode?

yes, updated the comment

ros-discourse:

ros-discourse:

PR #205, opened 2025-09-06
Add dual session tests

ShravanDeva5327:
Continuation of #191 

Tests for both `Trace` launch action and `ros2 trace` command are added with `dual_session` enabled

### DIFF
--- a/test_ros2trace/test/test_ros2trace/test_trace.py
+++ b/test_ros2trace/test/test_ros2trace/test_trace.py
@@ -18,7 +18,9 @@ import os
 import shutil
 import subprocess
 import tempfile
+import threading
 import time
+from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -30,6 +32,12 @@ from launch import LaunchService
 from launch_ros.actions import Node
 from lttngpy import impl as lttngpy
 import osrf_pycommon.process_utils
+import rclpy
+from rclpy.clock import Clock, ClockType
+from rclpy.duration import Duration
+from rclpy.executors import SingleThreadedExecutor
+from rclpy.qos import QoSProfile
+from std_msgs.msg import String
 from tracetools_launch.action import Trace
 from tracetools_read import get_event_name
 from tracetools_test.mark_process import get_corresponding_trace_test_events
@@ -272,6 +280,19 @@ class TestROS2TraceCLI(unittest.TestCase):
         process = self.run_command(['ros2', 'trace', *args], env=env)
         return self.wait_and_print_command_output(process)
 
+    def wait_for(
+        condition: Callable[[], bool],
+        timeout: Duration,
+        sleep_time: float = 0.1,
+    ):
+        clock = Clock(clock_type=ClockType.STEADY_TIME)
+        start = clock.now()
+        while not condition():
+            if clock.now() - start > timeout:
+                return False
+            time.sleep(sleep_time)
+        return True
+
     def run_nodes(self, env: Optional[Mapping[str, str]] = None) -> None:
         # Set trace test ID env var for spawned processes
         additional_env: Dict[str, str] = {}
@@ -339,9 +360,6 @@ class TestROS2TraceCLI(unittest.TestCase):
 
         loop = osrf_pycommon.process_utils.get_loop()
         launch_task = loop.create_task(ls.run_async())
-
-        # Give the launch service time to start actions
-        loop.run_until_complete(asyncio.sleep(1.0))
 
         return ls, launch_task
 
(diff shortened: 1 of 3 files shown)


ShravanDeva5327:
@christophebedard can you please review this?

test_tracetools_launch/test/test_tracetools_launch/test_trace_action.py
```diff
@@ -378,6 +413,7 @@ def test_action_substitutions_frontend_xml(self) -> None:
             <launch>
                 <arg name="session-name" default="my-session-name" />
                 <arg name="snapshot-mode" default="false" />
+                <arg name="dual-session" default="false" />
```
christophebedard, line 416: This (and `snapshop-mode`, which I didn't notice in #195) should be used below with `<trace>`, e.g.:

```
snapshot-mode="$(var snapshot-mode)"
dual-session="$(var dual-session)"
```

Same for the YAML launch file in `test_action_substitutions_frontend_yaml` below.

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -736,6 +786,97 @@ def test_snapshot_start_pause_resume_stop(self) -> None:
...
         shutil.rmtree(tmpdir)
 
+    @unittest.skipIf(not are_tracepoints_included(), 'tracepoints are required')
+    def test_dual_session_start_pause_resume_stop(self) -> None:
+        tmpdir = self.create_test_tmpdir('test_dual_session_start_pause_resume_stop')
+        session_name = 'test_dual_session_start_pause_resume_stop'
+
+        # Pre-configure dual session, start nodes and check that the snapshot session exists
+        ls, launch_task = self.pre_configure_dual_session_and_run_nodes(tmpdir, session_name)
+        self.assertTracingSessionExist(session_name+'-snapshot', snapshot_mode=True)
```
christophebedard: Tests often have more hardcoded values/literals (e.g., strings, numbers) because it makes test code more explicit and easier to read, but I think in this case we should have variables for the snapshot & runtime session names, since they're used so many times in this test:

```py
session_name_snapshot = session_name + '-snapshot'
session_name_runtime = session_name + '-runtime'
```

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -736,6 +786,97 @@ def test_snapshot_start_pause_resume_stop(self) -> None:
...
+            ]
+        )
+        self.assertEqual(0, ret)
+        trace_dir = os.path.join(tmpdir, session_name)
+        snapshot_trace_dir = os.path.join(tmpdir, session_name, 'snapshot')
+        runtime_trace_dir = os.path.join(tmpdir, session_name, 'runtime')
+        self.assertTraceExist(snapshot_trace_dir)
+        self.assertTraceExist(runtime_trace_dir)
+        self.assertTracingSessionExist(session_name+'-snapshot', snapshot_mode=True)
+        self.assertTracingSessionExist(session_name+'-runtime')
```
christophebedard: We should check here (and not after pausing below) that the _snapshot_ trace contains initialization events/data. You can simply use the list with `topic_name` -> `/ping`/`/pong` below.

----

However, this brings a potential problem (which doesn't mean we shouldn't do the above!). We first start the launch file, which configures the snapshot session and starts the nodes, and then we trigger a snapshot here. This creates a race: If we end up triggering the snapshot before the nodes have had time to properly start, the snapshot trace will not contain the expected initialization events. You added a 1-second sleep, but we've seen in nightly CI jobs that nodes can be really slow to start, so we try to avoid relying on hardcoded sleeps. The more robust solution here is to wait for the publishers/subscriptions (`/ping`, `/pong`) to exist before running `ros2 start` here.

To do that, you can use `rclpy` (which is already an indirect dependency of `test_ros2trace`, but you'll need to make it an explicit `<test_depend>` in the `package.xml`) to:

1. Create a node (e.g., using this test's name).
2. Create a `/ping` subscription and a `/pong` subscription to match the publishers created by `test_ping` & `test_pong`.
3. Use `Subscription.get_publisher_count()` to wait for `>= 1` publisher for each of the 2 subscriptions. You'll have to write some logic that waits for `sub.get_publisher_count() >= 1` and sleeps just a bit (e.g., 01 second) in between checking for publishers, but eventually times out if it takes too long, e.g., more than 10 seconds. Some similar code: https://github.com/ros2/rosbag2/blob/311921bf488018aa3d05d7f042bd6f5038338c2a/rosbag2_py/test/common.py#L43-L54 used for example here https://github.com/ros2/rosbag2/blob/311921bf488018aa3d05d7f042bd6f5038338c2a/rosbag2_py/test/test_transport.py#L114-L117.

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -736,6 +786,97 @@ def test_snapshot_start_pause_resume_stop(self) -> None:
...
+        # Pause tracing and check trace
+        ret = self.run_trace_subcommand(['pause', session_name, '--dual-session'])
+        self.assertEqual(0, ret)
+        self.assertTracingSessionExist(session_name+'-snapshot', snapshot_mode=True)
+        self.assertTracingSessionExist(session_name+'-runtime')
+        expected_trace_data = [
+            ('topic_name', '/ping'),
+            ('topic_name', '/pong'),
+        ]
+        num_events = self.assertTraceContains(trace_dir, expected_field_value=expected_trace_data)
```
christophebedard, line 865: This should check the _runtime_ trace data specifically as opposed to the whole trace, just to make sure the runtime trace contains what we think it should contain.

That means you can't use `topic_name` here, since that's from an init tracepoint (that wouldn't be recorded by the runtime trace, since it starts recording too late), and runtime events don't contain a lot of strings like topic names. I _think_ you could instead check the trace event names using [`_name`](https://github.com/ros2/ros2_tracing/blob/e5ad7f251b9836b99c03811606983107b6467d50/tracetools_read/tracetools_read/__init__.py#L51):

```py
        expected_trace_data = [
            ('_name', 'ros2:rcl_publish'),
            ('_name', 'ros2:rmw_publish'),
        ]
```

test_ros2trace/test/test_ros2trace/test_trace.py
```diff
@@ -736,6 +786,97 @@ def test_snapshot_start_pause_resume_stop(self) -> None:
...
+            trace_dir=tmpdir,
+            expected_field_value=expected_trace_data,
+        )
+        self.assertEqual(num_events, new_num_events, 'unexpected new events in trace')
+
+        # Resume tracing
+        ret = self.run_trace_subcommand(['resume', session_name, '--dual-session'])
+        self.assertEqual(0, ret)
+        self.assertTracingSessionExist(session_name+'-snapshot', snapshot_mode=True)
+        self.assertTracingSessionExist(session_name+'-runtime')
```
christophebedard: I'd like to check that resuming with `--dual-session` triggers a new snapshot. However, you can't really check the trace data for that new/second snapshot, since it's probably empty because there are no init events later at runtime. You can simply check that there's 1 directory under `${session_name}/snapshot/` before resuming and 2 directories after resuming.

For example, I just tested it:

```console
$ ll ~/.ros/tracing/my-dual-tracing-session/snapshot/
total 16
drwxrwx--- 4 christophe.bedard christophe.bedard 4096 Sep  7 12:21 ./
drwxrwx--- 4 christophe.bedard christophe.bedard 4096 Sep  7 12:20 ../
drwxrwx--- 3 christophe.bedard christophe.bedard 4096 Sep  7 12:20 snapshot-20250907-122029-0/
drwxrwx--- 3 christophe.bedard christophe.bedard 4096 Sep  7 12:21 snapshot-20250907-122110-1/
```

christophebedard (changes requested):
Looks good, I just have some minor-ish requests!

christophebedard:
Note: Shravan will pick this back up after finishing the GSoC final submission/report.

ros-discourse:

ros-discourse:

christophebedard:
As mentioned offline, I will give it another shot next week.

Issue #25, opened 2022-11-22
Provide example of real-time benchmark/use of ros2_tracing


christophebedard:
(Issue moved from GitLab, see: https://gitlab.com/ros-tracing/ros2_tracing/-/issues/131)

christophebedard:
This issue is very old, and the goal isn't very clear. Real-time use of `ros2_tracing` is mostly up to the kernel configuration and the LTTng configuration, which currently can't be done through `ros2_tracing`. I'll just close this.

PR #235, opened 2026-02-18
Clarify sub-buffer configuration support in Trace action (#25)

SarthakBagga:
## Description

This PR updates the README to clarify that sub-buffer sizes can be configured via the `Trace` launch file action (`subbuffer_size_ust` and `subbuffer_size_kernel`).

It also clarifies that other parameters, such as the read timer interval and explicit sub-buffer count, are not currently exposed through the `Trace` action or the `ros2 trace` command.

## Motivation

Improve the accuracy and clarity of documentation around trace configuration capabilities.

## Is this user-facing behavior change?

No (documentation clarification only).

## Did you use Generative AI?

Yes.

## Additional Information

All other previously proposed changes have been removed to keep the scope focused on this documentation clarification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,7 +336,7 @@ However, some settings need to be tuned for it to be fully real-time safe and fo
     * usually done the first time a tracepoint is executed within a thread for URCU thread registration, but registration can be manually performed to force it to be done during your application's initialization
     * see [this LTTng mailing list message](https://lists.lttng.org/pipermail/lttng-dev/2019-November/029409.html)
 
-[^rt-1]: this setting cannot currently be set through the [`Trace` launch file action](#launch-file-trace-action) or the [`ros2 trace` command](#trace-command), see [#20](https://github.com/ros2/ros2_tracing/issues/20)
+[^rt-1]: sub-buffer sizes are configurable via the [`Trace` launch file action](#launch-file-trace-action) (`subbuffer_size_ust` and `subbuffer_size_kernel`). Other settings such as the read timer interval and explicit sub-buffer count are not currently exposed through the `Trace` action or the [`ros2 trace` command](#trace-command), see [#20](https://github.com/ros2/ros2_tracing/issues/20)
 
 For further reading:
 


SarthakBagga:
Happy to make any changes or improvements if needed.

SarthakBagga:
can someone please review this and tell if any changes are needed

README.md
```diff
@@ -336,7 +336,7 @@ However, some settings need to be tuned for it to be fully real-time safe and fo
     * usually done the first time a tracepoint is executed within a thread for URCU thread registration, but registration can be manually performed to force it to be done during your application's initialization
     * see [this LTTng mailing list message](https://lists.lttng.org/pipermail/lttng-dev/2019-November/029409.html)
 
-[^rt-1]: this setting cannot currently be set through the [`Trace` launch file action](#launch-file-trace-action) or the [`ros2 trace` command](#trace-command), see [#20](https://github.com/ros2/ros2_tracing/issues/20)
+[^rt-1]: sub-buffer sizes are configurable via the [`Trace` launch file action](#launch-file-trace-action) (`subbuffer_size_ust` and `subbuffer_size_kernel`). Other settings such as the read timer interval and explicit sub-buffer count are not currently exposed through the `Trace` action or the [`ros2 trace` command](#trace-command), see [#20](https://github.com/ros2/ros2_tracing/issues/20)
```
christophebedard, line 339: Updating this section is a great idea, thank you. Let's keep this, revert the other changes, and update the PR description/title.

christophebedard (changes requested):
Thank you for the PR.

Unfortunately, I don't think we should take this. The original issue is very old, and the original goal wasn't very clear or well-defined. I'll close that issue. The launch file added here doesn't really do any real-time tracing. What's important for real-time is how the kernel and LTTng are configured, and we can't do that through `ros2_tracing` currently, like the note in the README says. However, the update to that note is great, so we can keep that.

In the future, for issues that aren't very well defined (especially ones that are old), I would highly recommend commenting on the issue to check with a maintainer before starting to work on it. This way, you can make sure the direction you're going to take is the right one.

Finally, this PR seems mostly AI-generated. If you do use AI, please mention what parts of the PR you used it for and confirm that you carefully validated its output.

SarthakBagga:
Thanks for the feedback!

I've updated the PR to keep only README clarifications and removed the other changes.
I’ll be more explicit about ai usage and more cautious with scope in future contributions.

Please let me know if anything else needs adjustment.

christophebedard (approved):
Thank you again for the PR!

I'll skip running ci.ros2.org since this only modifies the README.

PR #236, opened 2026-02-25
Set default values on TraceTestCase to avoid errors on >=8.2.0 pytest

claraberendsen:
## Description

In pytest >=8.2.0 the unittest  logic has been reworked to work more like a Function/Class, (see pytest-dev/pytest/pull/12089). This has the side-effect that during the collection phase all test are instantiated and given that the __init__ func of TraceTestCase includes required keyword arguments not being fulfilled in the collection phase it errors.

This  is avoided by setting default values on the __init__ definition so a `TraceTestCase.__init__() ` run during collection phase won't error.

pytest changelog: https://docs.pytest.org/en/stable/changelog.html#pytest-8-2-0-2024-04-27

Another approach to solve this would be to set `__test__=false` at the start to prevent pytest from collecting and instantiating the test, however that would need `__test__=true` for all of the subclasses of `TraceTestCase` that are actual tests for them to continue working. This seemed more cumbersome than this solution. 
 

Fixes regressions on resolute 

Partially, even though default values are set were they would error if not present, this was already the case for two out of the four default keywords implicitly. It won't fail now if those arguments are not set, as it would before for the ones that had no implicit default value. 

https://github.com/ros2/ros2_tracing/blob/757d1a9649c3eb0115d22005e220399cf540c1af/tracetools_test/tracetools_test/case.py#L80-L82

No
### Additional Information

Reference build: [Build Status](https://ci.ros2.org/job/ci_linux/27803/)

Build with fixes: [Build Status](https://ci.ros2.org/job/ci_linux/27805/)

Note that both have failed for reasons different to this PR, but in the later one there are 61 less regressions.


christophebedard (approved):
Seems OK to me with green CI if this doesn't affect the actual tests in any way (i.e., the value of these constructor arguments).

tracetools_test/tracetools_test/case.py
```diff
@@ -63,10 +63,10 @@ class TraceTestCase(unittest.TestCase):
...
         self,
         *args,
-        session_name_prefix: str,
-        events_ros: List[str],
-        package: str,
-        nodes: List[str],
+        session_name_prefix: str = '',
+        events_ros: List[str] = [],
+        package: str = '',
+        nodes: List[str] = [],
```
christophebedard: although I don't think it has any effect here, lists as default values can be dangerous:

```suggestion
        session_name_prefix: str = '',
        events_ros: Optional[List[str]] = None,
        package: str = '',
        nodes: Optional[List[str]] = None,
```

then below you can just do `events_ros or []` (maybe in parentheses due to the concatenation with another list?) and `nodes or []`.

claraberendsen:
Pulls: ros2/ros2_tracing#236
Gist: https://gist.githubusercontent.com/claraberendsen/685e45f6ed27e02cdfab5a0e3ba0e73a/raw/d757847a7f509b8f97fdfd8e42c0a3f72f1ce397/ros2.repos
BUILD args:  --packages-above-and-dependencies tracetools_test
TEST args:  --packages-above tracetools_test
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18329
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27831/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20848/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7697/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26817/)

PR #241, opened 2026-03-31
Loaned message tests

### LINKED ISSUE #240: Add loaned message pub-sub test to test_tracetools
### Description

https://github.com/ros2/rmw_cyclonedds/pull/566 and https://github.com/ros2/rmw_fastrtps/pull/868 added the `rmw_take` tracepoint to the loaned message take functions.

We should add a [test for this to `test_tracetools`](https://github.com/ros2/ros2_tracing/tree/rolling/test_tracetools/test). It would be very similar to `test_pub_sub.py` but with loaned messages. We specifically want to check that there is a `rmw_take` trace event for the loaned message received/taken by the subscription node.

### Motivation

Better test coverage == better code

### Design / Implementation Considerations

Implement test first, then try running the test without the tracepoints from the PRs mentioned above. The test should fail, indicating that the test nodes indeed do use loaned messages. Otherwise the nodes might incorrectly use normal/non-loaned messages and we might never find out.

### Additional Information

_No response_

nightduck:
## Description

This provides unit tests for the tracepoints added in [rmw_cyclonedds PR#556](https://github.com/ros2/rmw_cyclonedds/pull/566) and [rmw_fastrtps PR#871](https://github.com/ros2/rmw_fastrtps/pull/871), confirming that the rmw_take tracepoint is called even when using zero-copy

Fixes #240 

No

Yes, the initial draft was written entirely in Cursor's agentic tool. Review and revisions were done with a human (except the _CYCLO_DDS_SHM_XML string, which was AI generated)

### DIFF
--- a/test_tracetools/src/test_ping_loaned.cpp
+++ b/test_tracetools/src/test_ping_loaned.cpp
@@ -43,12 +43,10 @@ public:
     sub_ = this->create_subscription<Msg>(
       SUB_TOPIC_NAME,
       rclcpp::QoS(QUEUE_DEPTH),
-      [this](const Msg & msg) {
-        this->callback(msg);
-      });
+      std::bind(&PingLoanedNode::callback, this, std::placeholders::_1));
     pub_ = this->create_publisher<Msg>(
       PUB_TOPIC_NAME,
-      rclcpp::QoS(QUEUE_DEPTH));
+      rclcpp::QoS(QUEUE_DEPTH).transient_local());
 
     if (!sub_->can_loan_messages()) {
       throw std::runtime_error(
@@ -64,9 +62,9 @@ public:
   : PingLoanedNode(options, true) {}
 
 private:
-  void callback(const Msg & msg)
+  void callback(const Msg::ConstSharedPtr msg)
   {
-    RCLCPP_INFO(this->get_logger(), "[output] %u", msg.data);
+    RCLCPP_INFO(this->get_logger(), "[output] %u", msg->data);
     if (do_only_one_) {
       rclcpp::shutdown();
     }
--- a/test_tracetools/src/test_ping_loaned.cpp
+++ b/test_tracetools/src/test_ping_loaned.cpp
@@ -16,6 +16,7 @@
 #include <chrono>
 #include <cstring>
 #include <memory>
+#include <stdexcept>
 
 #include "rclcpp/rclcpp.hpp"
 #include "std_msgs/msg/u_int32.hpp"
@@ -48,6 +49,12 @@ public:
     pub_ = this->create_publisher<Msg>(
       PUB_TOPIC_NAME,
       rclcpp::QoS(QUEUE_DEPTH).transient_local());
+
+    if (!sub_->can_loan_messages() || !pub_->can_loan_messages()) {
+      throw std::runtime_error(
+              "message loaning is not available (publisher/subscription cannot loan messages)");
+    }
+
     timer_ = this->create_wall_timer(
       500ms,
       std::bind(&PingLoanedNode::timer_callback, this));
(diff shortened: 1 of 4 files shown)


nightduck:
I locally confirmed that this passes when built against the latest version of [rmw_cyclonedds](https://github.com/ros2/rmw_cyclonedds/tree/1cf18c066e5c9cc6c4069b68c4b242155d606d2f) and [rmw_fastrtps PR#871](https://github.com/ros2/rmw_fastrtps/pull/871), and it fails when built against versions of rmw_cyclonedds and rmw_fastrtps from before these tracepoints were added:
rmw_cyclonedds: [e7f44399fbadb474f0136370c232ed7eaf5be581](https://github.com/ros2/rmw_cyclonedds/tree/e7f44399fbadb474f0136370c232ed7eaf5be581)
rmw_fastrtps: [e5c8a84386fd9059f99ed6710ad4d7ac80f089b0](https://github.com/nightduck/rmw_fastrtps/tree/0fda8bbaa0d075222c8cdc5812c1c1fa94b1d83a)

test_tracetools/src/test_ping_loaned.cpp
```diff
@@ -0,0 +1,116 @@
...
+    timer_ = this->create_wall_timer(
+      500ms,
+      std::bind(&PingLoanedNode::timer_callback, this));
+  }
+
+  explicit PingLoanedNode(rclcpp::NodeOptions options)
+  : PingLoanedNode(options, true) {}
+
+private:
+  void callback(const Msg & msg)
```
christophebedard: shouldn't the callbacks accept a `ConstSharedPtr` for proper loaning?

test_tracetools/src/test_ping_loaned.cpp
```diff
@@ -0,0 +1,116 @@
...
+  {
+    sub_ = this->create_subscription<Msg>(
+      SUB_TOPIC_NAME,
+      rclcpp::QoS(QUEUE_DEPTH),
+      [this](const Msg & msg) {
+        this->callback(msg);
+      });
+    pub_ = this->create_publisher<Msg>(
+      PUB_TOPIC_NAME,
+      rclcpp::QoS(QUEUE_DEPTH));
```
christophebedard: any reason to not make both publishers transient local, like the existing ping/pong test nodes?

test_tracetools/src/test_ping_loaned.cpp
christophebedard: I guess what matters is what we take the loan from `rmw` (even if later it gets copied since we have a normal `const Msg &` callback), but I'd still try to use a proper loan-compatible callback

test_tracetools/src/test_ping_loaned.cpp
nightduck: Only the ping pub and pong sub are transient, but I'll add those changes to match

test_tracetools/src/test_ping_loaned.cpp
nightduck: I agree. Even if not important for the tests themselves, unit tests should stand out as examples of proper styling and usage

christophebedard (approved):

christophebedard:
Pulls: ros2/ros2_tracing#241
Gist: https://gist.githubusercontent.com/christophebedard/da8cd515281e71dc599c8b9a52e9add7/raw/dfc312de713dd967b3efd6ada33f92f258a077d7/ros2.repos
BUILD args:  --packages-above-and-dependencies test_tracetools
TEST args:  --packages-above test_tracetools
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/19000
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28764/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21583/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8606/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27631/)

christophebedard:
@nightduck the new tests are failing on all Linux jobs. Can you take a look? Also, GitHub CI doesn't seem to like launching roudi.

PR #249, opened 2026-05-05
Bump GitHub CI to Ubuntu 26.04 Resolute

christophebedard:
## Description

Rolling has migrated to Ubuntu 26.04 Resolute. Also, bump some action versions.

Also, explicitly install babeltrace CLI for Resolute source sanity check job. lttng-tools gets installed by rosdep because it's a dependency of tracetools. In [24.04 and older](https://packages.ubuntu.com/noble/lttng-tools), it also depends on babeltrace (the CLI), but [on 26.04](https://packages.ubuntu.com/resolute/lttng-tools), it switched to depending on babeltrace2. We need to switch to babeltrace2, but for now just explicitly install babeltrace in the source sanity check job.

### Additional Information

Backport to Lyrical. We will separately need to change "rolling" to "lyrical" in the CI config matrices for it.

### DIFF
--- a/.github/workflows/dco.yml
+++ b/.github/workflows/dco.yml
@@ -5,9 +5,9 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v6
     - name: Set up Python 3.x
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: '3.x'
     - name: Check DCO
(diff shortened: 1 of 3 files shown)


christophebedard:
This only changes the local GitHub CI config, so we can skip ROS 2 CI.

christophebedard:
There's a build issue with Fast DDS related to `-Wmaybe-uninitialized` that I can't figure out. It seems to be due to a recent commit on the 3.6.x branch. I'm curious to see if it will show up in the next ci.ros2.org nightlies; I'd proceed with this for now.

cottsay (approved):

christophebedard:
@Mergifyio backport lyrical

PR #181, opened 2025-05-30
[kilted] Update CI config for Kilted

christophebedard:
## Description

Update the CI config to properly test against Kilted.

### Is this user-facing behavior change?

No, just internal GitHub CI.

### Did you use Generative AI?

No.

PR #251, opened 2026-05-12
[lyrical] Update CI config for Lyrical

christophebedard:
## Description

Similar to #181

The GitHub CI config was switched to Ubuntu Resolute in #249 / #250. Now switch to `lyrical`.


ahcorde (approved):

christophebedard:
I will skip ci.ros2.org since this only modifies the local GitHub CI config

PR #256, opened 2026-06-26
Add rosidl buffer tracing points

nvcyc:
## Description

Fixes # (issue)


tracetools/include/tracetools/tracetools.h
```diff
@@ -264,6 +264,59 @@ _DECLARE_TRACEPOINT(
...
+ * \param[in] cpu_subscriber_count matched CPU-only buffer subscribers
+ * \param[in] accel_endpoint_count matched accelerated buffer endpoints
+ * \param[in] selected_route selected route (`buffer` or `legacy_fallback`)
+ * \param[in] reason route reason
+ */
+_DECLARE_TRACEPOINT(
+  rmw_buffer_publish_route,
+  const void * rmw_publisher_handle,
+  const void * message,
+  const char * topic_name,
```
christophebedard, line 286: here and everywhere else you use `char * topic_name`: You don't need to collect the topic name. We can cross-reference it using the `rmw_publisher_handle`. If you look at the design doc (and/or the paper), this is done everywhere here to minimize the data we collect, especially at runtime.

tracetools/include/tracetools/tracetools.h
```diff
@@ -264,6 +264,59 @@ _DECLARE_TRACEPOINT(
...
+ * Buffer-aware publish route decision.
+ *
+ * \param[in] rmw_publisher_handle pointer to the publisher's `rmw_publisher_t` handle
+ * \param[in] message pointer to the ROS message being published
+ * \param[in] topic_name topic name
+ * \param[in] total_matched total matched subscriber count
+ * \param[in] buffer_aware_count matched subscribers known as buffer-aware
+ * \param[in] cpu_subscriber_count matched CPU-only buffer subscribers
+ * \param[in] accel_endpoint_count matched accelerated buffer endpoints
+ * \param[in] selected_route selected route (`buffer` or `legacy_fallback`)
```
christophebedard, line 279: if `selected_route` is binary (buffer/legacy_fallback), a simple `bool` would be more optimal

tracetools/include/tracetools/tracetools.h
```diff
@@ -264,6 +264,59 @@ _DECLARE_TRACEPOINT(
...
+  rmw_buffer_publish_route,
+  const void * rmw_publisher_handle,
+  const void * message,
+  const char * topic_name,
+  const size_t total_matched,
+  const size_t buffer_aware_count,
+  const size_t cpu_subscriber_count,
+  const size_t accel_endpoint_count,
+  const char * selected_route,
+  const char * reason)
```
christophebedard, line 292: here and everywhere else you use `char * reason`: Do you really need a C string here? If it doesn't offer much more information beyond what one can infer from `selected_route`, it might be better to leave it out. Otherwise, you should use enums: https://lttng.org/man/3/lttng-ust/v2.14/#doc-tracepoint-enum

christophebedard (commented):
I know this is just a draft, but since you're adding a lot of tracepoints, it would be important to:

1. Add a/some sections to the design doc to cover this new set of tracepoints and how they're supposed to work together at a high level: https://github.com/ros2/ros2_tracing/blob/rolling/doc/design_ros_2.md
2. Mention roughly what each tracepoint is supposed to do or collect (i.e., link what values across relevant tracepoints) in `tracetools.h` (see the doc for other tracepoints there)
3. Add tests: https://github.com/ros2/ros2_tracing/tree/rolling/test_tracetools/test

I've also left some comments.

PR #262, opened 2026-08-19
Isolate tracing tests with private LTTng homes

### LINKED ISSUE #99: Fix LTTng issue with `test_ros2trace` and `test_tracetools_launch` so that they can run tracing tests in parallel
This is a follow-up to #94.

#95 changed the tracing tests to be able to run them in parallel with normal tests and tracing tests. However, there was still an issue when `test_ros2trace` and `test_tracetools_launch` were running tests in parallel in ci.ros2.org, see: https://github.com/ros2/ros2_tracing/pull/95#issuecomment-1995314441. To temporarily work around it, we made `test_ros2trace` depend on `test_tracetools_launch` so that they wouldn't run tests in parallel, see: https://github.com/ros2/ros2_tracing/pull/95#issuecomment-2008123673.

Now, we need to fix the issue and remove the workaround.

sylvesterkaczmarek:
## Description

Remove the serialization workaround between `test_ros2trace` and `test_tracetools_launch` by giving each package its own LTTng control state.

- add `tracetools_test.isolated_lttng_home()` to create a private `LTTNG_HOME`
- use session-scoped autouse fixtures in both tracing-test packages
- stop only the private session daemon during teardown, restore the previous environment, and remove the temporary home
- remove the false `test_ros2trace` dependency on `test_tracetools_launch`

This allows the build farm to schedule the two packages in parallel without sharing the same LTTng session daemon.

Fixes #99

Yes. AI was used to assist with tests.

### Additional Information

Testing performed:

- `PYTHONPATH=tracetools_test pytest -q tracetools_test/test/test_lttng.py` → 3 passed
- Python syntax checks for the added helper and fixtures
- package XML parsing
- `git diff --check`

### DIFF
--- a/test_ros2trace/package.xml
+++ b/test_ros2trace/package.xml
@@ -29,12 +29,6 @@
   <test_depend>tracetools_test</test_depend>
   <test_depend>tracetools_trace</test_depend>
 
-  <!-- TODO(clalancette): This is actually a false dependency,
-	  but we add it here to ensure that we don't run multiple
-	  tracing tests in parallel, which often ends up crashing.
-  -->
-  <test_depend>test_tracetools_launch</test_depend>
-
   <export>
     <build_type>ament_python</build_type>
   </export>
--- a/test_ros2trace/test/conftest.py
+++ b/test_ros2trace/test/conftest.py
@@ -1,0 +1,25 @@
+# Copyright 2026 Sylvester Kaczmarek
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Iterator
+
+import pytest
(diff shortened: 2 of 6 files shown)


fujitatomoya:
@sylvesterkaczmarek same here can you comply our AI guideline to disclose the information with using our PR template? and DCO is missing.

sylvesterkaczmarek:
@fujitatomoya Done, thanks. DCO is corrected and the PR description now follows the ROS 2 template.

fujitatomoya:
>PYTHONPATH=tracetools_test pytest -q tracetools_test/test/test_lttng.py → 3 passed

this does not add up. why only 3 tests? can you share what test command did you execute and result?

fujitatomoya:
@sylvesterkaczmarek btw, i saw you created other PRs, those do not comply the issue template and disclose of AI use.

sylvesterkaczmarek:
Thanks for flagging this. The `3 passed` line referred only to the new helper unit file, not the whole tracing test stack. The exact local command was:

`PYTHONPATH=tracetools_test pytest -q tracetools_test/test/test_lttng.py` -> `3 passed`.

I rechecked the current head (`e43d618`): the changed helper and both package fixtures pass `py_compile`, and both modified `package.xml` files parse successfully. More importantly, the repository CI has now run the full Ubuntu 26.04 / Rolling matrix on this head and all jobs are green, including source and binary builds plus tracing tests with instrumentation enabled/disabled and tracepoints included/excluded.

I do not have a local Linux/LTTng environment, so I am not claiming a local full integration run. The earlier wording was too narrow; the full package-level validation here is the repository CI matrix.

